PR #1, opened 2016-06-06
Initial pr

dawallin:
Here is the first pull request so we can get something to discuss around.


bensigelman:
nit (while looking at the file list): is NuGet.exe meant to be a part of the PR?

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace OpenTracing
+{
+    public interface ISpan<T> 
+    {
+        T GetSpanContext();
+
+        void Finish();
+        void FinishWithOptions(DateTime finishTime);
+
+        void SetBaggageItem(string restrictedKey, string value);
```
bensigelman: unlike SetTag and Log, SetBaggageItem is meant to be paired with a GetBaggageItem...

also, I am fairly confident we are going to remove restrictions on the keys.

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace OpenTracing
+{
+    public interface ISpan<T>
```
bensigelman: fine to leave them out for now, but before merging we will want some clear comments for everything in ISpan and ITracer.

OpenTracing/OpenTracing.OpenTracing/ITracer.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace OpenTracing
+{
+    public interface ITracer<T>
+    {
+        ISpan<T> StartSpan(string operationName);
+        ISpan<T> StartSpan(StartSpanOptions startSpanOptions);
+
+        void Inject(ISpan<T> span, IInjectCarrier<T> carrier);
+        bool TryJoin(string operationName, IExtractCarrier<T> carrier, out ISpan<T> span);
```
bensigelman: I'm surprised about the `out ISpan<T> span`: why not have `TryJoin` return that type instead? It also seems like there will be public/private problems trying to populate `out`, though maybe I just don't know C# well enough...

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System;
+
+namespace OpenTracing
+{
+    public interface ISpan<T> 
+    {
+        T GetSpanContext();
```
bensigelman: This (and the templating on a `T` context type) is the most controversial thing in this PR, as @yurishkuro pointed out. Without getting into a discussion (yet) of whether it's better or worse, are you able to quickly describe how things would be different if you got rid of the `<T>` and the `GetSpanContext()`?

dawallin:
I'm looking at other OSS projects and sometimes I found it included. It is the artifact package builder. I change and exclude it for now and need to look up the best practice for this.

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
dawallin: My thought to get the baggage was to do something like this.
```csharp
var memoryCarrier = new MemoryTextMapCarrier(new Dictionary<string, string>() { });
tracer.Inject(span, memoryCarrier);            
var value = memoryCarrier.TextMap[key]
```
But I agree, a GetBaggage(key, value) method on the span is more convinient. I will add it.

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
dawallin: Yes, I'll add this.

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
dawallin: This is changed in the next coming update.

OpenTracing/OpenTracing.OpenTracing/ISpan.cs
bensigelman: Thanks...

Also, a general point of order: I was inspired by your `SpanContext` stuff and am attempting to find a compromise where other languages can incorporate a context concept at the OpenTracing API layer without introducing tons of new terminology and cognitive load for the user. I will keep you informed of the RFC PR on that front which I hope to send out this weekend. (I've done the interesting part but need to clean things up)

bensigelman:
@dawallin: somewhat related to the context stuff, see https://gitter.im/opentracing/public?at=575c92732eaa837d71e8a0fd

cwe1ss:
closed, because we merged #2

PR #2, opened 2016-08-21
improvements for PR #1

cwe1ss:
Hi guys!
I came across this project yesterday and really like the idea, so I took Daniel's PR and tried to continue with it. That's what I've done:

General:
* The library now targets netstandard1.1, which means it runs on almost all .NET platforms (full .NET v4.5+, .NET Core, Mono, Xamarin, Windows Phone, Universal Windows Platform).
* I changed the project structure to better align with current repositories from Microsoft (e.g. ASP.NET Core) (src/test folder, .project.json, ...). 
* psake based build script which runs the tests and creates the nuget packages. 
* There's no documentation in code yet - I would add that if you want to proceed with it.

API:

* I tried to get as close to the specification as possible. I renamed stuff (e.g. "message" -> "eventName", ...)
* I removed the generics. Implementations will want to have typed versions of `IScopeContext` but IMO having the generics uglifies the API for end users and casting in the implementations is easy. 
* I changed `DateTime` usages to `DateTimeOffset` and used UTC everywhere. `DateTime.Now` returns the local machine timezone and can't be used for such a library. `DataTimeOffset` has in implicit cast from `DateTime` so it also accepts them and handles timezones correctly.
* `ISpan`
  * removed `GetTracer()` - not in the spec and not necessary IMO
  * `SetTag` now accepts `object` as the value. There's no base class for numeric types in C# and having separate overloads for each type would break all implementations whenever a new type should be added. I think it's better to leave this open and let the implementation decide whether the type is OK or not. An alternative would be to limit it to `struct` types (a separate overload for string would still be required though).
  * Features that only extend the interface but don't require different implementations are moved into c# extension methods. This includes `SetBaggageItem` (which is a shortcut for `Context.SetBaggageItem()` and typed methods for all built in tags (e.g. `SetTagComponent`)
* `ISpanContext` is now a separate interface
* `ITracer` no longer has `TryJoin` - instead it has `Extract` as defined in the spec.
  * Extension methods for common use cases of `StartSpan` are in `TracerExtensions.cs`
* Propagation (Carriers/Formats): 
  * `IInjectCarrier` and `IExtractCarrier` are now only marker interfaces and the different implementations are wrapper-objects around the carrier data structure (e.g. text-map, http-headers, byte array). Implementers have to decide based on this type how to inject/extract.
  * Every format defines its own extension methods for easier use (user doesn't have to manually wrap the carrier object)
* There's now a No-op tracer called `NullTracer` - `Null*` is more common in the .NET world than `Noop*`

OpenTracing.Carrier.HttpClient: 
* This project is no longer required. The logic has been integrated into the `OpenTracing` package (`Propagation\HttpHeadersCarrier.cs`). I also changed the logic to no longer use the HttpClient as the carrier but instead the `HttpHeaders` dictionary (which is part of `HttpRequestMessage` and `HttpResponseMessage`). Using `HttpClient.DefaultRequestHeaders` was wrong because `HttpClient` instances are often shared between requests. Therefore, the headers must be set for every individual request.

OpenTracing.BasicTracer / OpenTracing.BasicTracer.Tests:
* I adjusted everything to the new API and made other changes as well - this is still pretty much work in progress.

Usage:

```csharp
// get tracer (would be through dependency injection in real apps)
ITracer tracer = GetTracer();

// create root span
var rootSpan = tracer.StartSpan("get_account")
    .SetBaggageItem("user-id", "1234")
    .SetTag("key", "value")
    .SetTagComponent("AspNetCore"); // intellisense for known tags

// create HTTP api request
var request = new HttpRequestMessage(HttpMethod.Post, "/some_api");
request.Content = new StringContent("{ \"userId\": 1234 }");

// Inject - typed extension methods for known carriers
tracer.InjectHttpHeaders(rootSpan, request.Headers);

// call api
var response = await new HttpClient().SendAsync(request);

// Extract new context from response (this time with the raw method)
var spanContext = tracer.Extract(Formats.HttpHeaders, new HttpHeadersCarrier(response.Headers));

// create new span with reference
var subSpan1 = tracer.StartSpan("sub-operation", SpanReference.ChildOf(spanContext))
    .SetTag("custom-key", "custom-value");

// create another follows_from span with StartOptions
var options = new StartSpanOptions()
    .FollowsFrom(rootSpan)
    .SetStartTimestamp(new DateTime(2016, 8, 21, 14, 0, 0, DateTimeKind.Utc));
var subSpan2 = tracer.StartSpan("sub-operation", options)
    .SetTagHttpUrl("http://example.com/api");

// finish operations
subSpan1.Finish();
subSpan2.Finish(new FinishSpanOptions()
    .SetFinishTimestamp(new DateTime(2016, 8, 21, 14, 0, 5, DateTimeKind.Utc)));
rootSpan.Finish();
```


cwe1ss:
I also just added an (untested) PoC for integration with ASP.NET Core which creates spans for every request and child-spans for every outgoing http-call.

cwe1ss:
There has been a long discussion about some of this on gitter with @yurishkuro - main discussion points:

* the extension methods for propagation (e.g. `InjectIntoHttpHeaders`) and recommended tags (e.g. `SetTagComponent`)
* propagation in general and the separate format parameter in Java ( https://github.com/opentracing/opentracing-java/issues/31 )

I still don't fully understand the format/carrier system so I'll dig deeper into the other libraries and come back with more questions :)

bensigelman:
@cwe1ss as an initial meta-comment: thank you for putting this together. I look forward to going through this in more detail. Should I/others hold off on that until you've grokked the history of the format/carrier debate, etc?

cwe1ss:
I think/hope, I now understand everything well enough. I updated everything and I think it's now good enough to look at it :)

The most important thing:

I don't yet have the typed constraint between format/carrier because C# has extension methods. They make it possible for format providers and for library vendors to extend the `ITracer` interface for users. For this reason, format is a simple string right now.

This means, the inject contract of `ITracer`: 

```csharp
void Inject(ISpanContext spanContext, string format, IInjectCarrier carrier);
```

can be extended with something like this: (see `src/OpenTracing/Extensions` for more examples)

```csharp
public static void InjectHttpHeaders(this ITracer tracer, ISpanContext spanContext, HttpHeaders headers)
{
    tracer.Inject(spanContext, Formats.HttpHeaders, new HttpHeadersCarrier(headers));
}
```

This makes it very easy to use because people don't even have to manually wrap the carrier object for most cases. It also means that you can't get the format wrong as long as you use the extension methods. 
If there are cases where the extension methods are not enough, you can still use the raw method.

There's some other stuff I'd like to discuss but it might be too much to do all of this in one github issue. So as soon as you give me a first general ok for the code, I would create some more issues with specific questions?!

src/OpenTracing.BasicTracer/Tracer.cs
```diff
@@ -46,29 +46,55 @@ public ISpan StartSpan(string operationName, StartSpanOptions options)
             return span;
         }
 
-        public void Inject(ISpanContext spanContext, IInjectCarrier carrier)
+        public void Inject(ISpanContext spanContext, string format, IInjectCarrier carrier)
```
yurishkuro: been a few years since I've done C#, but can't the `format` argument be a typed constant? E.g.

```
    public interface Format<T> { }
    
    public class StringFormat<T>: Format<T> {
        private string code;
        public StringFormat(string code) {
            this.code = code;
        }
    }
    
    public class Builtin {
        // using `int` as carrier type here
        public static StringFormat<int> TextMap = new StringFormat<int>("buildin.textmap");
    }
```

cwe1ss:
@dawallin great to hear that you are ok with using my PR. I added some of your stuff from #3 to my PR - hope that's ok for you as well?! 

> During the summer we had a long discussion about immutable ISpanContext. We decided that Get and Set baggage would be on the ISpan and ISpanContext would only have a GetAll method.

I removed the methods from ISpanContext!

> For the StartSpanOptions and the SpanBuilder I was inspired by the Java implementation (but removed the abstract and introduced it as an extension method instead). I agree that the API is very simple but I think it makes the creation of a new span even more fluent

I agree that this looks a little bit better. I added your code and removed my methods on `StartSpanOptions`. The only difference now is that going through the Builder requires an additional object - but I guess that's not a huge problem?!
However, I kept the (IMO) most used methods as StartSpan extension methods on ITracer - they don't go through the builder. Maybe that's a good compromise. Take a look at it!

@bensigelman @yurishkuro: As disussed with @dawallin, it would be great if you could merge this PR so that we have common baseline to continue from. We should discuss all necessary changes in separate issues so that @dawallin and I can coordinate our efforts.

dawallin:
Nice work @cwe1ss . I'm still struggling with my Visual Studio to load and run the project but it LGTM. I'm fine to start with this and target the remaining in separate issues.

cwe1ss:
I recently do must of the library work in Visual Studio Code - it starts fast and is good enough for this kind of work (intellisense, syntax highlighting, ...)!

dawallin:
Thanks, I will take a deeper look at it. I use VS Code to look at the other opentracing implementations, but haven't used it for C# yet.

src/OpenTracing.AspNetCore/OpenTracingHttpMessageHandler.cs
```diff
@@ -0,0 +1,62 @@
+using System;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+
+namespace OpenTracing.AspNetCore
+{
+    public class OpenTracingHttpMessageHandler : DelegatingHandler
```
yurishkuro: is this going to be compiled into the same assembly as the API itself?

src/OpenTracing.AspNetCore/OpenTracingHttpMessageHandler.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            var httpContext = _httpContextAccessor.HttpContext;
+            var tracer = (ITracer)httpContext.RequestServices.GetService(typeof(ITracer));
+
+            ISpan requestSpan = (ISpan)httpContext.Items[typeof(ISpan)];
+            ISpan callSpan = null;
+
+            try
+            {
+                // TODO operationName?
```
yurishkuro: yeah, always a pain in std instrumentations

src/OpenTracing.AspNetCore/OpenTracingMiddleware.cs
```diff
@@ -0,0 +1,65 @@
...
+            var existingSpanContext = tracer.ExtractHttpHeaders(context);
+            ISpan requestSpan = null;
+
+            try
+            {
+                // TODO peer tags?
+
+                requestSpan = tracer.StartSpan(context.Request.Path, SpanReference.ChildOf(existingSpanContext))
+                    .SetTagComponent("AspNetCore")
+                    .SetTagSpanKindServer()
```
yurishkuro: problem here - a Zipkin-compatible tracer needs to know `span.kind=server` as part of the `StartSpan` call, not after, because StartSpan ends up creating a child span with a fixed span ID, and RPC server span has to use the same span ID as the incoming spanContext (Zipkin's restriction), as opposed to a normal child (e.g for outbound call) which gets a new span ID. For that reason StartSpan in other languages allows passing tags ahead of the actual span creation.

src/OpenTracing.BasicTracer/Baggage.cs
```diff
@@ -0,0 +1,66 @@
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+
+namespace OpenTracing.BasicTracer
+{
+    public class Baggage
+    {
+        /// <summary>
+        /// <para>Baggage keys have a restricted format: implementations may wish to use them
```
yurishkuro: there are no restrictions anymore

src/OpenTracing/ISpan.cs
```diff
@@ -0,0 +1,84 @@
...
+namespace OpenTracing
+{
+    /// <summary>
+    /// Span represents an active, un-finished span in the OpenTracing system.
+    /// Spans are created by the <see cref="ITracer"/> interface.
+    /// </summary>
+    public interface ISpan : IDisposable
+    {
+        // TODO IDisposable behavior?
+        // TODO SetOperationName() ?
```
yurishkuro: yes

src/OpenTracing/ITracer.cs
```diff
@@ -0,0 +1,31 @@
...
+        ISpan StartSpan(string operationName, StartSpanOptions options = null);
+
+        /// <summary>
+        /// This takes the <paramref name="spanContext"/> and injects it for propagation within <paramref name="carrier"/>. 
+        /// The actual type of <paramref name="carrier"/> depends on the value of <paramref name="format"/>.
+        /// </summary>
+        /// <param name="spanContext">The <see cref="ISpanContext"/> instance to inject into the carrier.</param>
+        /// <param name="format">The format which should be used.</param>
+        /// <param name="carrier">See the documentation for the chosen <paramref name="format"/> for a description of the carrier object.</param>
+        void Inject(ISpanContext spanContext, string format, IInjectCarrier carrier);
```
yurishkuro: https://github.com/opentracing/opentracing-csharp/pull/2#discussion-diff-75928451

src/OpenTracing/Propagation/TextMapCarrier.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            if (payload == null)
+            {
+                throw new ArgumentNullException(nameof(payload));
+            }
+
+            _payload = payload;
+        }
+
+        public IEnumerable<KeyValuePair<string, string>> GetEntries()
```
yurishkuro: do these need some sort of "override" annotations?

src/OpenTracing.AspNetCore/AspNetCoreFormatExtensions.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+using Microsoft.AspNetCore.Http;
+using OpenTracing.AspNetCore;
+using OpenTracing.Propagation;
+
+namespace OpenTracing
+{
+    public static class AspNetCoreFormatExtensions
+    {
+        public static ISpanContext ExtractHttpHeaders(this ITracer tracer, HttpContext context)
```
yurishkuro: is `HttpContext` a standard thing in .NET? Neat.

yurishkuro:
I took a pass through, not very thorough as there are too many files (would be better to have separate PRs for BasicTracer, extensions, etc.)

One thought on the numerous extensions: in other languages we erred on the side of providing fewer helper methods until we actually get more integrations and experience with the API.

Also, it would be nice to have separation of the compiled artifacts into separate assemblies, so that the actual API dependencies could be minimal (maybe it already is, I cannot read the project files).

And finally, I think we're missing out on language capabilities to enforce matching types between format and injectors/extractors by using generics. You can see the corresponding issue in Java that introduced the Format, the original version had similar separation of inject vs. extract carriers, and generics were able to capture the correct type for each operation for given format (but for various reasons Java went with merged carrier type and one-arg generic `Format<C>` - wasn't my preference).

src/OpenTracing.AspNetCore/OpenTracingHttpMessageHandler.cs
cwe1ss: no, OpenTracing.AspNetCore is just a proof of concept because I needed something to actually use the library.
I can remove it now or later - since this needs a lot more work, it should be in a separate repository anyway. Would you be interested in hosting something like this under the official opentracing github-project or should this be separate?

src/OpenTracing.AspNetCore/OpenTracingMiddleware.cs
cwe1ss: see comment above - AspNetCore is just a PoC and will be moved elsewhere (to be defined)

src/OpenTracing/Propagation/TextMapCarrier.cs
cwe1ss: no - interface implementations don't need that in C#, only derived classes.

src/OpenTracing.AspNetCore/AspNetCoreFormatExtensions.cs
cwe1ss: it's a standard in the ASP.NET Core framework. this is one of the reasons why I built this PoC, it shows the great potential of extension methods with `Inject`/`Extract` - framework implementations can provide their extension for whatever format is appropriate and users don't need to worry about wrapping the actual carrier into a opentracing-carrier.

cwe1ss:
> I took a pass through, not very thorough as there are too many files (would be better to have separate PRs for BasicTracer, extensions, etc.)

thank you very much for taking the time! It's much appreciated! We will definitely work with smaller and more focused PRs once we have a baseline merged. Don't expect this to be "final" already. I definitely want to create separate issues to discuss a few things after this has been merged.

> Also, it would be nice to have separation of the compiled artifacts into separate assemblies, so that the actual API dependencies could be minimal (maybe it already is, I cannot read the project files).

API dependencies already are at a minimum. `System.Net.Http` is part of the full .NET framework installation package so everyone has this out of the box. In .NET core you could theoretically not depend on it, but to keep .NET Core and full .NET consistent, Microsoft also created a "default meta package" for .NET core which includes it. 
This means, the extension methods in the main project are only for stuff that's included in the base class library anyway.

> And finally, I think we're missing out on language capabilities to enforce matching types between format and injectors/extractors by using generics. 

If it's ok for you, I'd also work on this, once the baseline is merged - this way it would be easier for Daniel and me to work on stuff together. We definitely need a good solution for formats/carriers!!

I'll be hiking in the mountains (Austria is beautiful :smile:) for the rest of the week so I will continue to work on this next week.

Thanks again for helping me with all of this!

src/OpenTracing.AspNetCore/OpenTracingHttpMessageHandler.cs
yurishkuro: opentracing-contrib sounds like the right place for extensions

dawallin:
How do we move on? Is there anything you (@bensigelman , @yurishkuro) think we need to do before this PR can be accepted? Would it be possible to accept this as a version 0.8-beta so that we can continue to work and solve the remaining problems in separate smaller PRs. Now when we are two contributors we need one master branch to make changes against. And to keep history of issues and discussions I think it is better here at opentracing than in any of our own repositories..

There are comments here that the repo contains to much outside the opentracing API. They are needed to test and proof the API and could in my opinion stay there until we reach version 1. If we don't want them here already now, could we create separate repositories for them, e.g. opentracing/basictracer-csharp and opentracing-contrib/csharp-AspNetCore.

bensigelman:
@dawallin @cwe1ss I will take a quick look now for big-picture stuff. Once we've resolved things at that level it would be great to move on to smaller incremental PRs.

I would be very enthusiastic about moving non-core-api code (basictracer, AspNetCore, etc) into other repos, though I don't feel super strongly about whether that happens before or after this PR merges.

bensigelman:
PS: thank you!

NuGet.config
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bensigelman: meta-comment: I'm just trusting you on what should be `.gitignore`d and what shouldn't be...

src/OpenTracing/ISpan.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <returns>The current <see cref="ISpan"/> instance for chaining.</returns>
+        ISpan SetTag(string key, object value);
+
+        /// <summary>
+        /// Records an event with optional payload data for this Span.
+        /// </summary>
+        /// <param name="eventName">Name of the event.</param>
+        /// <param name="payload">An optional payload object.</param>
+        /// <returns>The current <see cref="ISpan"/> instance for chaining.</returns>
+        ISpan LogEvent(string eventName, object payload = null);
```
bensigelman: just FYI, logging is a bit of a moving target right now... I'm going to make some progress on https://github.com/opentracing/opentracing.github.io/issues/96 today or this weekend.

src/OpenTracing/ISpan.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace OpenTracing
+{
+    /// <summary>
+    /// Span represents an active, un-finished span in the OpenTracing system.
+    /// Spans are created by the <see cref="ITracer"/> interface.
+    /// </summary>
+    public interface ISpan : IDisposable
+    {
+        // TODO IDisposable behavior?
```
bensigelman: (I don't know what this is)

src/OpenTracing/ITracer.cs
bensigelman: (would be interested to read a response to @yurishkuro's comment, here or where he linked to)

src/OpenTracing/NullTracer/NullSpan.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    public class NullSpan : ISpan
+    {
+        internal static readonly NullSpan Instance = new NullSpan(NullSpanContext.Instance);
+
+        public ISpanContext Context { get; }
+
+        private NullSpan(ISpanContext context)
+        {
+            Context = context;
```
bensigelman: why store this? why not just have `Context` above return `NullSpanContext.Instance`?

src/OpenTracing/NullTracer/NullSpanContext.cs
```diff
@@ -0,0 +1,20 @@
+using System.Collections.Generic;
+
+namespace OpenTracing.NullTracer
+{
+    public class NullSpanContext : ISpanContext
+    {
+        internal static readonly NullSpanContext Instance = new NullSpanContext();
+
+        private readonly Dictionary<string, string> _baggage = new Dictionary<string, string>();
```
bensigelman: I guess there's no `EMPTY_SET` like in Java?

src/OpenTracing/Propagation/BinaryCarrier.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+
+namespace OpenTracing.Propagation
+{
+    /// <summary>
+    /// A <see cref="ITextMapCarrier"/> which allows a byte array to be used as a carrier object.
+    /// </summary>
+    public class BinaryCarrier : IInjectCarrier, IExtractCarrier
```
bensigelman: This is a matter of taste, but I think it's fine to collapse `IInjectCarrier` and `IExtractCarrier` into plain-old `ICarrier`... slightly less flexible/clean, but slightly fewer concepts for new users.

src/OpenTracing/Propagation/HttpHeadersCarrier.cs
```diff
@@ -0,0 +1,55 @@
+using System;
+using System.Collections.Generic;
+using System.Net.Http.Headers;
```
bensigelman: Is this truly ubiquitous in C#? If not, should we do something more abstract/flexible here? Other OT impls don't *force* the use of a specific HTTP header type.

src/OpenTracing/Propagation/HttpHeadersCarrier.cs
```diff
@@ -0,0 +1,55 @@
...
+{
+    /// <summary>
+    /// A <see cref="ITextMapCarrier"/> which allows <see cref="HttpHeaders"/> implementations to be used as carrier objects.
+    /// </summary>
+    /// <remarks>
+    /// <see cref="HttpHeaders"/> is a multi-value dictionary. Since most other platforms represent http headers as regular
+    /// dictionaries, this carrier represents it as a regular dictionary to tracer implementations.</remarks>
+    public class HttpHeadersCarrier : ITextMapCarrier
+    {
+        // TODO should this class be internal/private? It's name is misleading because this only works with the .NET framework class "HttpHeaders"
```
bensigelman: ah, I just saw this comment; this should only be part of the core OT API if.f. the .NET HttpHeaders is nearly ubiquitous.

src/OpenTracing/SpanBuilder.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace OpenTracing
+{
+    /// <summary>
+    /// Helper builder for convenience to construct startoptions and create a new span.
+    /// </summary>
+    public class SpanBuilder
```
bensigelman: this is an interface in Java and I think that's the right thing for the core layer... esp for the noop impl.

src/OpenTracing/SpanReference.cs
```diff
@@ -0,0 +1,83 @@
...
+        public static SpanReference ChildOf(ISpanContext parent)
+        {
+            return parent != null ? new SpanReference(TypeChildOf, parent) : null;
+        }
+
+        /// <summary>
+        /// Creates a new "follows_from" reference for the <see cref="ISpanContext"/> of the given Span.
+        /// </summary>
+        /// <param name="parent">The parent Span.</param>
+        public static SpanReference FollowsFrom(ISpan parent) => FollowsFrom(parent?.Context);
```
bensigelman: I don't think we need as many places where we convert from Spans to SpanContexts... should be able to do that only in SpanBuilder and nowhere else, I think.

src/OpenTracing/StartSpanOptions.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+
+namespace OpenTracing
+{
+    /// <summary>
+    /// Provides a way to configure the behavior of <see cref="ITracer.StartSpan"/>.
+    /// </summary>
+    public class StartSpanOptions
```
bensigelman: (we can get rid of this if Builder is an interface)

bensigelman:
EOF for my comments... Basically looks really good overall, thanks again again.

src/OpenTracing/ISpan.cs
dawallin: IDisposible makes it possible to create a scope with using and force the program to call Finish when the scope is finished.
```
using(tracer.StartSpan())
{
   //Do something with the span.
} // Finish on the span is called automatically, even if an exception is thrown.
```
It is mainly used for managed resources, like files that needs to be closed. I think it is called `AutoCloseable` in Java.

src/OpenTracing/Propagation/BinaryCarrier.cs
dawallin: I like them separated. For example in dot net there is a http client in the standard library for  which we need a carrier implementing the inject interface. There are different web servers that need there own carriers just implementing the extract interface. Combining the interfaces would just force non applicable methods throwing NotImplementedExceptions.

src/OpenTracing/StartSpanOptions.cs
dawallin: I don't think you should compare SpanBuilder to its Java version. In Java the way to create a span is through the buildSpan method. 
Here the method on the tracer interface is the StartSpan(OperationName, StartSpanOptions) method. This makes it very clear for the implementor. For the user this felt cumbersome if you have many start options. SpanBuilder is a helper extension for building the StartSpan parameters in a fluent API. There is no logic i the SpanBuilder. 
There are comments above that it could be postponed until we know more about the API usage, or extracted to an external extension library, but I feel that it is very simple and simplifies the user code and could be left here for now..

src/OpenTracing/SpanReference.cs
dawallin: See comment on StartSpanOption.

src/OpenTracing/SpanBuilder.cs
dawallin: See comment on StartSpanOption.

src/OpenTracing/ISpan.cs
bensigelman: 👍

src/OpenTracing/StartSpanOptions.cs
bensigelman: I'm not understanding why we want the core OpenTracing API to support both the builder and the StartSpanOptions... As with any API, it's easier to add things than to subtract them. If these two approaches are semantically equivalent, let's choose the one that's easier for the caller (the builder IMO) and let implementations create a StartSpanOptions or similar if that makes their lives easier internally.

Forgive me if there's something that the Options or Builder approach can do that the other cannot...

dawallin: I see your point. Agree that the API footprint should be minimized. If anything should be cut out it is the SpanBuilder. Compared to the Java version the SpanBuilder is more of a SpanStartOptionBuilder, no logic, just a fluent way to add options. After collecting all the parameters instead of building the SpanStartOption like normal builder, it directly calls Tracer.StartSpan. I see it more like the predefined tags, it is just a convenient method to help guide the user. If it is stopping this PR we cut it out.

dawallin:
@bensigelman Is there anything from the comments above that we need to fix for this PR to be accepted?

bensigelman:
@dawallin I don't feel strongly about whether it happens before or after merging this PR, but I do think we should remove API surface area around SpanBuilder and/or StartSpanOptions before encouraging people to create dependencies on this code.

On that note, here are pointers to how the startSpan calling code looks in other languages:

Java: https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/Tracer.java#L33

Go: https://github.com/opentracing/opentracing-go/blob/master/tracer.go#L30

Python: https://github.com/opentracing/opentracing-python/blob/master/opentracing/tracer.py#L58

Javascript: https://github.com/opentracing/opentracing-javascript/blob/master/src/tracer.js#L33

It's important that the calling code has an easy way to adjust zero or more of the startSpan params without specifying all of them, and in that sense I think the SpanBuilder might actually be the part of the API we'd want to keep.

Getting back to your question: I'd say "let's merge this", but please file issues about ^^^ as well as the fate of the helper/example implementations and so on.

It would be good to see if @yurishkuro has any blocking concerns, too.

yurishkuro:
> It would be good to see if @yurishkuro has any blocking concerns, too.

Nope, since this isn't final.

NuGet.config
cwe1ss: NuGet is the .NET package manager (like NPM for node). This file defines which servers should be included for the package restore project.

As long as the file only contains this one default endpoint, it would not be necessary. But I like having it in the repo because it also has an explicit "clear" statement. This means, the repo still takes the default server, even if someone has a machine-level override.

src/OpenTracing/NullTracer/NullSpanContext.cs
cwe1ss: There's `Enumerable.Empty<Type>()` for read-only lists (which returns a cached object), but there's no equivalent for dictionaries (maps) because they are mutable.

cwe1ss:
I added `SetOperationName(string operationName)` to `ISpan`. I will create separate issues for the other topics!

src/OpenTracing/Propagation/Formats.cs
```diff
@@ -8,15 +11,15 @@ public static class Formats
         /// byte array instance. <see cref="ITracer.Inject"/> must append to the byte array carrier
         /// (rather than replace its contents).</para>
         /// </summary>
-        public const string Binary = "binary";
+        public static readonly Format<byte[]> Binary = new Format<byte[]>("binary");
```
yurishkuro: is byte array the right carrier for binary? I assume byte array has a fixed length, so especially for the injector how does the instrumentation know how long the tracing context in binary encoding will be? I don't know the native IO types in C#, but I would assume some sort of a byte buffer or i/o stream to be a better carrier than byte array.

src/OpenTracing/ITracer.cs
```diff
@@ -18,14 +18,14 @@ public interface ITracer
         /// <param name="spanContext">The <see cref="ISpanContext"/> instance to inject into the carrier.</param>
         /// <param name="format">The format which should be used.</param>
         /// <param name="carrier">See the documentation for the chosen <paramref name="format"/> for a description of the carrier object.</param>
-        void Inject(ISpanContext spanContext, string format, IInjectCarrier carrier);
+        void Inject<TCarrier>(ISpanContext spanContext, Format<TCarrier> format, TCarrier carrier);
```
yurishkuro: 👍

cwe1ss:
As of e9c8ae8, the format/carrier handling is now almost equal to the Java version. 

* Format is typed and there is a static accessor class called `Formats` which has `Formats.TextMap`, `Formats.HttpHeaders`, etc. 
  * The `Format` class has a `Name` property which is the actual factor for equality. I really don't like the Java version which only compares the actual object address.
* IInjectCarrier, IExtractCarrier is no longer required (Formats.Binary accepts a regular byte[] )
* I still have all the extension methods. I still really like them because IMO they make the API *a lot* easier to use for users.

I'm still not sure whether this is better or worse than our previous version :smile: but I guess it's more important to be equal to Java and solve any possible issues together when they arise.

One thing I'm worried about is when we have a case where a format should accept multiple incompatible carriers. E.g. now I use `ITextMap` for `Formats.HttpHeaders` - but as I've said, in .NET, http headers are typically represented as multi-value maps. We can now no longer define that .NET based trace implementations should be able to handle both versions (which would be better for performance most probably). We would have to introduce a new format instead which would be confusing for users.

Would be great if someone could merge this now. :smile: You could also give me contributor rights to this repository if you want - I would not do any non-trivial commits without a PR/your approval.

src/OpenTracing/Propagation/Formats.cs
cwe1ss: I'll look this up. Probably not.

dawallin:
I still have some problems with the format/carrier problem, but I will address that once this PR is accepted. 
Basically I think format and carrier are totally connected and therefore you only need one parameter, the other one is just over specification. What the generics do here is just forcing you that they are correct at compile time. I have a proposal for a different solution in my PR #3, I will try to move that concept to this branch once this is merged.

dawallin:
We have now added some issues and made some changes according to the major comments. @bensigelman Is there anything more before you can accepting this PR ?

bensigelman:
@dawallin nope! I will just give you both write privs in this repo, you certainly have more C# knowledge than I.

bensigelman:
(done)

cwe1ss:
awesome - thanks!

Issue #5, opened 2016-08-29
SpanBuilder vs StartSpanOptions vs StartSpan-parameters


cwe1ss:
As of #2, I'd like to discuss how we should pass parameters (tags, references, timestamp) to `ITracer.StartSpan`. There are multiple options for this:

### 1 - Optional parameters

```csharp
/* ITracer interface */

ISpan StartSpan(
  string operationName, /* only this is required, the rest is optional */
  DateTimeOffset? startTimestamp = null,
  IEnumerable<SpanReference> references = null,
  IDictionary<string, object> tags = null
);

/* Second method for single reference (because this is the most-used case) */
// this allows for easier consumer code and for tracer perf optimizations
// (in case they want to store single references without a list)
ISpan StartSpan(
  string operationName,
  DateTimeOffset? startTimestamp = null,
  SpanReference reference = null,
  IDictionary<string, object> tags = null
);

/* usage */

tracer.StartSpan("get_account"); // no optional parameters

tracer.StartSpan("get_account", startTimestamp: DateTime.UtcNow); // custom timestamp

// just one reference - using the second method
tracer.StartSpan("get_account", reference: SpanReference.ChildOf(someSpanContext)); 

// tags from a variable and reference list
var myTags = new Dictionary<string, object> { { "key", "value" }, { "key2", "value2"} };
tracer.StartSpan("get_account",
  references: new List<SpanReference> { ref1, ref2 },
  tags: myTags
);
```

**Advantages:**
* It doesn't create unnecessary objects (StartSpanOptions, SpanBuilder)
* Probably the best performance (because it requires less garbage collection)
* Extending the interface with additional optional parameters
  * doesn't break consumers
  * breaks trace implementations with a compile-time error (which is a good thing IMO because they now know that they have to implement it)
 
**Disadvantages:**
* Adding tags always requires creating the dictionary instance
* It's not possible to add typed extensions for known tags/references
* the method signature might get long if we add more parameters in the future

-----------

### 2 - StartSpanOptions

```csharp
/* ITracer interface */
ISpan StartSpan(string operationName, StartSpanOptions options);

/* StartSpanOptions class */
class StartSpanOptions {
  IList<SpanReference> References { get; set; } = new List<SpanReference>();
  IDictionary<string, object> Tags { get; set; } = new Dictionary<string, object>();
  DateTimeOffset? StartTimestamp { get; set; }

  // helper methods for nicer API
  StartSpanOptions AddReference(SpanReference reference);
[...]

PR #8, opened 2016-08-31
SpanBuilder interface for #5

cwe1ss:
As an example for #5, this would introduce a `ISpanBuilder` interface and remove the existing `StartSpanOptions` and the existing `SpanBuilder` implementation. 

This is exactly how the Java version is implemented.

I favor having the interface over an actual implementation to let trace implementations optimize the behavior without having users to update their OpenTracing package.


cwe1ss:
I will merge this on Tuesday If there are no complaints by then.

dawallin:
As I commented in #5 I prefer the StartSpanOptions, and having the builder as an optional extension for a fluent API style. But if you think the SpanBuilder is better I won't object.

If we go for the SpanBuilder then this PR looks perfectly fine.

cwe1ss:
Ok, I'll merge it!

Btw, it would be possible to go the other way and offer non-breaking extension methods like this one in the future:

```csharp
public static ISpan StartSpan(this ITracer tracer, string operationName, SpanReference reference)
{
    return tracer.BuildSpan(operationName).AddReference(reference).Start();
}

// Usage
tracer.StartSpan("get_account", SpanReference.ChildOf(parent));

// But it would even be more characters than the current solution
// so there's probably not much point... :-)
tracer.BuildSpan("get_account).AsChildOf(parent).Start();
```

PR #10, opened 2016-09-08
Generic carrier

dawallin:
I tried to improve the ITracer API with a generic Carrier to get rid of the format parameter at inject and extract. The format is instead added as a generic type on the Carrier. I think this uses the generics better than the existing solution, especially when we need to select a mapper to map between Format and SpanContext. I also think it makes the Formats easier, because it is independent of the carrier. The TextMapFormat is now a simple dictionary and HttpHeaderFormat is a dictionary with some restrictions on the keys. At last it removed the need of format extension methods to simplify the inject and extract call.

All the generic types are inferred from the parameters and the client user never need to care. 

While updating the basic tracer I fixed a few things that I think differed from other OT implementations.
- tracerId from GUID -> ulong
- sampled from random -> spanId % sampleRate
- basicOpenTracing propagation prefix changed to "ot-tracer-" and "ot-baggage-"
- introduced parentId on spanContext

@cwe1ss Comments are appreciated.


cwe1ss:
I think this no longer follows the spec. You now have "formats" and "carriers" holding data. Basically, your "format" is now already a "carrier" - but by inheriting from `Dictionary<string, string>` it creates a copy of the object instead of just holding a reference. IIUC, it's also no longer possible to reuse a carrier for multiple formats which is supported in the other languages.  

My main concern is that this is pretty complex and completely different to what Java does. Since almost all innovation in this area is coming from other languages, this will make it really hard for tracing systems to port their existing implementations to C#.

@yurishkuro Your thoughts?

PS: The following test fails because you just [replace your internal reference](https://github.com/dawallin/opentracing-csharp-1/blob/bb6f1234ef7840f1ab604d64cb27b79309e29c33/src/OpenTracing/Propagation/DictionaryCarrier.cs#L28) and never update the original object:

```csharp
var dictionary = new Dictionary<string, string>();
var dictionaryCarrier = new DictionaryCarrier(dictionary);

tracer.Inject(span.Context, dictionaryCarrier);

Assert.Equal(traceId.ToString(), dictionary["ot-tracer-traceid"]);
```

Your [existing test](https://github.com/dawallin/opentracing-csharp-1/blob/bb6f1234ef7840f1ab604d64cb27b79309e29c33/test/OpenTracing.BasicTracer.Tests/BasicTracerIntegrationTests.cs#L31) has a carrier that doesn't point to an external object. I'm not sure if this is a valid use case because IIUC you always want to inject the baggage data into a non-opentracing object.

yurishkuro:
I am not very familiar with C#'s implementation of generics. For example, in Java the generics are erased at runtime, so `IInjectCarrier<TFormat>` parameter is the same as `IInjectCarrier<Object>`, i.e. the tracer has no way of getting access to the `TFormat` part at runtime.

Overall, I agree with @cwe1ss comment- this is a radically different API from everything we have in other languages, and pretty much the only argument for the difference is "it is easier to use", contrary to any empirical proof that the current `(format, carrier)` API used in other languages is actually "difficult to use".  I would much prefer to have a uniform API across languages, covered by a single language-neutral spec, especially in the area that was historically the most difficult for people to grasp.

cc @bensigelman

cwe1ss:
Interesting - in C#, generics are a runtime feature also, that's why you can't have a list of types with different generics (like in the LogField<?> case).

bensigelman:
I have trouble assessing this due to my unfamiliarity with C#. The default reaction having looked at so many of these OT APIs in other languages is some sort of nearly unconscious "different is bad" risk assessment. That said, if C# generics are more "real" than Java's (i.e., more like C++'s or something), I can certainly imagine a world where the format becomes just a compile-time type specifier and nothing more. (This doesn't work in Java, of course, hence the explicit format arg)

I was hoping to spectate a bit as the people who actually know C# lay out more detailed arguments. All I can really do is express the obvious fact that this deviates from the norm in OT. Whether that's unacceptable is another question, but it's certainly a drawback that would have to be accompanied by a significant benefit in the C# universe.

cwe1ss:
@dawallin @yurishkuro How should we proceed with this? I'd like to come to a consensus within the next few days to make sure we can extract the BasicTracer etc.

To me this PR seems too different/complex and also currently not correct (see failing test from above) so I would vote for not merging it.

I think, if we want to get rid of the separate format parameter in the current solution, we have to discuss a different question: Is it really required to reuse carrier types for different formats? If every format had its own base interface, we could just switch responsibilities: Now the format defines the carrier, but then the carrier would define the format. In that case, trace implementations could just check the types. They would have to honor inheritance (for custom derived types) but this could easily be cached so there wouldn't be any performance hit. 

However, this also has some disadvantages and most of it was discussed in https://github.com/opentracing/opentracing-java/issues/31 - so I don't think we should open this discussion again for now.

dawallin:
We can close this not to block other issues. I see that I do it a different way and agree there is a win in having all implementations language-neutral . Have been thinking about it for some days and can't come up with a better solution even if I don't like how it is now. Thanks for the link, I missed that discussion before.

@cwe1ss Just for curiosity, You claim that I have errors in the unit tests but I get no errors when I run on my computer. Can you run it again? Is it still failing?

cwe1ss:
@dawallin Sorry, I probably wasn't clear. You don't have a test for this scenario right now, but if you add a new test with the following code from my initial comment, it will fail:

```csharp
var dictionary = new Dictionary<string, string>();
var dictionaryCarrier = new DictionaryCarrier(dictionary);

tracer.Inject(span.Context, dictionaryCarrier);

Assert.Equal(traceId.ToString(), dictionary["ot-tracer-traceid"]);
```

dawallin:
@cwe1ss Isn't that exactly this test [DefaultBasicTracer_WhenSpanInjectedToMemoryCarrier_Work](https://github.com/dawallin/opentracing-csharp-1/blob/ef1ba17ed30cf64e2da8aa3841c69ec4514a82a1/test/OpenTracing.BasicTracer.Tests/BasicTracerIntegrationTests.cs#L31). It passes.

cwe1ss:
@dawallin nope, you are not using an external dictionary. you are just setting the values on the `MemoryTextMapCarrier` itself. I think it shouldn't even be possible to use the MemoryTextMapCarrier without passing an external dictionary into it. AFAIU you always want to inject your span context into some opentracing-external object (e.g. a `System.Net.Http.Headers.HttpHeaders` dictionary).

Your code which succeeds:
```csharp
var dictionaryCarrier = new MemoryTextMapCarrier();
tracer.Inject(span.Context, dictionaryCarrier);
Assert.Equal(traceId.ToString(), dictionaryCarrier.TextMap["ot-tracer-traceid"]);
```
If you change it to this, it fails:
```csharp
var dictionary = new Dictionary<string, string>();
tracer.Inject(span.Context, new MemoryTextMapCarrier(dictionaryCarrier));
Assert.Equal(traceId.ToString(), dictionary["ot-tracer-traceid"]);
```
The problem is [in this line](https://github.com/dawallin/opentracing-csharp-1/blob/ef1ba17ed30cf64e2da8aa3841c69ec4514a82a1/src/OpenTracing/Propagation/MemoryTextMapCarrier.cs#L28) where you replace your member variable instead of copying each value back into the original dictionary. This way, the original dictionary never changes.

PR #11, opened 2016-09-09
DateTime instead of DateTimeOffset; resolves #4

### LINKED ISSUE #4: DateTime vs DateTimeOffset vs long (ticks) for timestamps
We have to decide whether we should use `DateTime`, `DateTimeOffset` or `long` (ticks) for timestamps. Unfortunately, it seems like all of them have certain disadvantages...

Interesting articles:
* [Choosing between DateTime, DateTimeOffset, TimeSpan, and TimeZoneInfo](https://msdn.microsoft.com/en-us/library/bb384267(v=vs.110).aspx)
* [Converting Times between Time Zones](https://msdn.microsoft.com/en-us/library/bb397769(v=vs.110).aspx)
* [Converting between DateTime and DateTimeOffset](https://msdn.microsoft.com/en-us/library/bb546101(v=vs.110).aspx)

Some important quotes:

> The `DateTime` structure is suitable for applications that do the following:
* ..., Work with UTC dates and times only, ...
* When saving or sharing DateTime data, UTC should be used and the DateTime value's Kind property should be set to DateTimeKind.Utc

> The `DateTimeOffset` type includes all of the functionality of the DateTime type along with time zone awareness. This makes it is suitable for applications that do the following:
* Uniquely and unambiguously identify a single point in time. The DateTimeOffset type can be used to unambiguously define the meaning of "now", to log transaction times, to log the times of system or application events, and to record file creation and modification times.
* Note: These uses for DateTimeOffset values are much more common than those for DateTime values. As a result, DateTimeOffset should be considered the default date and time type for application development.

Converting:
> For UTC and local DateTime values, the Offset property of the resulting DateTimeOffset value accurately reflects the UTC or local time zone offset. 

> However, for DateTime values whose Kind property is DateTimeKind.Unspecified, these two conversion methods produce a DateTimeOffset value whose offset is that of the local time zone.

> Ifyour application requires that converted DateTime values unambiguously identify a single point in time, you should consider using the DateTimeOffset.UtcDateTime property to handle all DateTimeOffset to DateTime conversions.

----------

This means, `DateTime` would be possible if we would only use UTC. However, for some reason many developers don't use `DateTime.UtcNow` and `DateTime.Now` properly. While the former correctly returns a UTC timestamp, the latter returns a timestamp for the current timezone on which the host is running - and this of course is wrong for server scenarios.

Since it's not guaranteed that everybody uses `UtcNow`, if we would use `DateTime`, every trace implementation would have to make sure that it either always calls `ToUniversalTime()` (or only if `Kind` is "Local"). there's also DateTimeKind.Unspecified which seems to be handled similar to local - this is the kind you get when you manually create a DateTime object.

`DateTimeOffset` seems to be more explicit about all of this, however it still uses the local server time zone when you pass a date with kind=Unspecified.

By using `DateTime` or `DateTimeOffset` we would need to define some recommendations - e.g. Users should always use UTC, tracer implementations should convert kind=local to UTC, tracer implementations should throw an exception if kind=Unspecified, ...

--------

Using `long ticks` would be an alternative which maybe would make it a bit more obvious for users that they can't just pass a regular `DateTime.Now` and instead have to think about the right thing to do. However, I guess you might still run into the same issues if you use `DateTime.Now.Ticks` or `new DateTime(...).Ticks` with an unspecified kind. I haven't yet tested this though.

Other drawbacks of using ticks are that the trace implementations now can't do any validation and it's not obvious for users and trace implementations, whether the ticks are UNIX epoch ticks or Windows ticks.

Your opinions?

cwe1ss:
As discussed in #4 we are using DateTime instead of DateTimeOffset for the following reasons:

* It doesn't automatically convert values with DateTimeKind.Local
* We can let trace implementations decide whether they want to convert non-UTC values.
* It's the most-known type for users
* It's a little bit faster
* We can switch to DateTimeOffset later without breaking consumers because there's an implicit cast. (we would only break trace implementations)

I added a recommendation about using `DateTimeKind.Utc` in the comments of each usage.

### DIFF
--- a/src/OpenTracing.BasicTracer/LogData.cs
+++ b/src/OpenTracing.BasicTracer/LogData.cs
@@ -4,14 +4,14 @@ namespace OpenTracing.BasicTracer
 {
     public struct LogData
     {
-        public LogData(DateTimeOffset timestamp, string eventName, object payload)
+        public LogData(DateTime timestamp, string eventName, object payload)
         {
             Timestamp = timestamp;
             EventName = eventName;
             Payload = payload;
         }
 
-        public DateTimeOffset Timestamp { get; private set; }
+        public DateTime Timestamp { get; private set; }
         public string EventName { get; private set; }
         public object Payload { get; private set; }
     }
--- a/src/OpenTracing.BasicTracer/Span.cs
+++ b/src/OpenTracing.BasicTracer/Span.cs
@@ -12,8 +12,8 @@ namespace OpenTracing.BasicTracer
         public ISpanContext Context => _context;
 
         public string OperationName { get; private set; }
-        public DateTimeOffset StartTimestamp { get; }
-        public DateTimeOffset? FinishTimestamp { get; private set; }
+        public DateTime StartTimestamp { get; }
+        public DateTime? FinishTimestamp { get; private set; }
 
         public IDictionary<string, object> Tags { get; } = new Dictionary<string, object>();
         public IList<LogData> Logs { get; } = new List<LogData>();
@@ -22,7 +22,7 @@ namespace OpenTracing.BasicTracer
             ISpanRecorder spanRecorder,
             SpanContext context,
             string operationName,
-            DateTimeOffset startTimestamp,
+            DateTime startTimestamp,
             IDictionary<string, object> tags)
         {
             if (spanRecorder == null)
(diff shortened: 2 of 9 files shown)


dawallin:
LGTM

PR #15, opened 2016-09-12
Naming: SpanTags->Tags

cwe1ss:
Renames to be closer to the Java version. Both names don't reflect common types in .NET so I think this should be fine and a bit more readable. `Field` from the Key-Value PR will also not contain a prefix.

I'll merge this if there are no complaints by Wednesday.


dawallin:
LGTM

cwe1ss:
SpanReference has been removed with #16 - this PR now only contains the rename from `SpanTags` to `Tags`.

cwe1ss:
Since I have some time today, I'll just merge this. I hope that's ok for everyone else - if not, please just let me know!

PR #17, opened 2016-09-13
Propagation improvements, Tests

cwe1ss:
This PR does the following things:

* `Format` is now a struct instead of a class. Value comparison for structs is done by comparing each field so this is less code.
* Adds a new Test project with tests for `Format` equality and `DictionaryCarrier` behavior. It's a new project because the one for BasicTracer will be moved to the other repository.
* `ITextMap.Add` was renamed to `Set` because the Java equivalent `Put` means that keys are replaced if they already exist. `Set` is the common name in C# for this.
* Renames `HttpHeadersCarrier` to `SystemNetHttpHeadersCarrier`. This makes clear that this is not the base class but instead for one actual headers type. Users will usually not see this because they call the extension methods (which still have to be discussed in a separate issue)


cwe1ss:
This doesn't introduce fundamental changes, so I'd merge this on Friday/Saturday if there are no complaints by then.

Note that there's a general issue regarding the Carrier implementation here: #24

dawallin:
This is good. Nice to get rid of the IEquatable. I think we can merge this independent of what is discussed in #24. 
Only objection is that I miss the changes in the solution file for the new test project.

cwe1ss:
> Only objection is that I miss the changes in the solution file for the new test project.

good catch - thank you, I've added the necessary changes.

I'll merge this now!

PR #18, opened 2016-09-13
TagExtensions for ISpanBuilder and ISpan

cwe1ss:
This PR should serve as a discussion point about whether or not we should use extension methods for known tags. Just as a recap: "Extension methods" in C# do not change the original contract of a type (`ISpanBuilder`, `ISpan` in this case). They are just syntactical sugar for users and they will be replaced with regular static calls by the compiler. In other words, the only reason for them is to offer a simpler/better experience for users.

This is different to what Java does because afaik Java does not have extension methods. I really like extension methods because they offer great Intellisense for users and people can easily extend it with their own extension methods to keep a consistent API.

Of course, they could also live in a separate library (e.g. opentracing-contrib) but since they are so fundamental, I don't like that basically everybody would have to add two libraries.

The current code already contains `SetTag*` extensions for `ISpan`, this PR adds the same extensions for `ISpanBuilder`.

Using extension methods for known tags would allow users do write the following code:

```csharp
// Using ISpanBuilder
var span = tracer.BuildSpan("get_account")
  .WithTagComponent("ASP.NET Core")
  .WithTagSpanKindServer()
  .WithTagHttpUrl(request.Path)
  .WithTagHttpMethod(request.Method)
  .WithTag("custom", "value")
  .Start();

// Using ISpanBuilder + ISpan
var span = tracer.BuildSpan("get_account")
  .WithTagSpanKindServer()
  .Start();

span
  .SetTagHttpUrl(request.Path)
  .SetTagHttpStatusCode(response.StatusCode)
  .SetTag("custom", "value");

// Syntax without extension methods
var span = tracer.BuildSpan("get_account")
  .WithTag(Tags.Component, "ASP.NET Core")
  .WithTag(Tags.SpanKind, Tags.SpanKindServer)
  .Start();

span
  .SetTag(Tags.HttpStatusCode, (int)response.StatusCode);
```

If you agree to keep them, my main question is:
* Should we remove "Tag" from the names? Example: Should it be `.SetComponent("ASP.NET Core")` instead of `SetTagComponent("ASP.NET Core")`?
  * The advantage is a more readable API
  * Disadvantage is that it might not be clear to the user that this will be saved as a tag (although that might not be important for the user anyway)
  * It could also lead to conflicts should we decide to add a regular `SetComponent` method in the future for whatever reason.

Open tasks:
- [ ] Decision: Should we use Tag extension methods at all?
- [ ] Should we keep "Tag" in the name?
- [ ] rename file `SpanTagExtensions.cs` to `TagExtensions.cs` (the diff would have been awful if I had done this in one step)

### DIFF
--- a/src/OpenTracing/Extensions/SpanTagExtensions.cs
+++ b/src/OpenTracing/Extensions/SpanTagExtensions.cs
@@ -10,7 +10,7 @@ namespace OpenTracing
     /// <summary>
     /// Typed extension methods for common tags.
     /// </summary>
-    public static class SpanTagExtensions
+    public static class TagExtensions
     {
         /// <summary>
         ///  "span.kind" hints at the relationship between spans, e.g. client/server.
@@ -23,9 +23,25 @@ namespace OpenTracing
         /// <summary>
         ///  "span.kind" hints at the relationship between spans, e.g. client/server.
         /// </summary>
+        public static ISpanBuilder WithTagSpanKind(this ISpanBuilder builder, string spanKind)
+        {
+            return WithTag(builder, Tags.SpanKind, spanKind);
+        }
+
+        /// <summary>
+        ///  "span.kind" hints at the relationship between spans, e.g. client/server.
+        /// </summary>
         public static ISpan SetTagSpanKindServer(this ISpan span)
         {
-            return SetTagSpanKind(span, "server");
+            return SetTagSpanKind(span, Tags.SpanKindServer);
+        }
+
+        /// <summary>
+        ///  "span.kind" hints at the relationship between spans, e.g. client/server.
+        /// </summary>
+        public static ISpanBuilder WithTagSpanKindServer(this ISpanBuilder builder)
+        {
+            return WithTagSpanKind(builder, Tags.SpanKindServer);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


cwe1ss:
@bensigelman @dawallin @yurishkuro ping :smile:

src/OpenTracing/Extensions/SpanTagExtensions.cs
```diff
@@ -23,17 +23,41 @@ public static ISpan SetTagSpanKind(this ISpan span, string spanKind)
...
         public static ISpan SetTagSpanKindClient(this ISpan span)
         {
-            return SetTagSpanKind(span, "client");
+            return SetTagSpanKind(span, Tags.SpanKindClient);
+        }
+
+        /// <summary>
+        ///  "span.kind" hints at the relationship between spans, e.g. client/server.
+        /// </summary>
+        public static ISpanBuilder WithTagSpanKindClient(this ISpanBuilder builder)
```
yurishkuro: to me WithClientSpanKindTag would've read better ` ¯\_(ツ)_/¯`. But a brute-force syntax would've read even better
```
builder.WithTag(Tags.SpanKind, Tags.SpanKindClient)
```
i.e. in this case the shortcut is worse than the verbose original

yurishkuro:
Don't know what to say. The only marginal benefit I see in this is strong typing, which I think can be achieved with generics instead, similar to `Format<C>` we can have `TypedTag<V>` and an extension method `WithTag(TypedTag<V> tag, V value)`. But the actual individual methods look overly cumbersome, I'd prefer `WithTag(Tag.XYS, value)` syntax, which is more explicit, easier to read (imo), and requires minimal number of extensions.

dawallin:
I don't have a strong opinion but I lean toward @yurishkuro suggestion that this does not add much extra value, and the raw `WithTag(Tag.X, value)` is easy enough to both write or read. 
I liked the suggestion with TypedTag to get the strong typing.

cwe1ss:
OK! I'll remove the extensions for tags!

cwe1ss:
removed in master with commit 791808f

Issue #21, opened 2016-09-13
Integrate with Travis build server


cwe1ss:
I have seen that other projects use travis-ci for build integration and I would like to set this up for this repo as well. 

* Who can create the build project in the opentracing account and link it with this repository?
* @dawallin do you have experience with travis config files for .NET? (It should be using the new `dotnet cli`). If not, I'll try to create a config file - I've only worked with AppVeyor so far so I'll probably need a few shots/commits to get it right. 

My plan is to have a script that runs build, tests and creates the nuget packages as artifacts. If travis also has some way to do deployments to nuget, that would be even better.

dawallin:
I have no experience of Travis. I integrated to AppVeyor before but they had problems with the new NUnit test runner so I never got a green light from the tests. Now when we use XUnit it maybe works better. I didn't had time to get all the nuget packaging and deploying configured so there is not much at AppVeyor today.

I have no preferences, but to me it looks like AppVeyor is more Windows centric and Travis is more Linux centric . If you have the time to look into Travis I would say go.

cwe1ss:
The repo now contains a .travis.yml file that should work (at least it did in my branch :smile:). 

Could someone please activate this repository in the opentracing travis account? @bensigelman @dawallin @yurishkuro

bensigelman:
@cwe1ss done!

![image](https://cloud.githubusercontent.com/assets/2251553/18528938/e8859b78-7a7f-11e6-9b83-822061915e1f.png)

bensigelman:
oh, and

![image](https://cloud.githubusercontent.com/assets/2251553/18528948/f5cdd318-7a7f-11e6-9a65-20a5dac1a394.png)

cwe1ss:
@bensigelman thank you!!

working build:
![travis-build](https://cloud.githubusercontent.com/assets/4581460/18529555/d87cc780-7acd-11e6-8038-cf81640cc376.PNG)

and there's now also the badge on the repo:
![travis-badge](https://cloud.githubusercontent.com/assets/4581460/18529589/fde128a4-7acd-11e6-8754-b08da0e3f6af.PNG)

PR #22, opened 2016-09-14
New build scripts from ASP.NET Core for windows & linux builds

cwe1ss:
for #21 

This uses the build scripts from the ASP.NET Core repositories (e.g. https://github.com/aspnet/logging ) and their build system called "KoreBuild". It's a bit of an overkill but it works and it results in successful build on travis.

The current psake/powershell solution didn't run on linux so I removed it. (Powershell is now open source but it's still alpha and I didn't find a working solution)

Important: build.sh must be executable: http://stackoverflow.com/questions/33820638/travis-yml-gradlew-permission-denied


cwe1ss:
Immediately merged because it's just infrastructure stuff.

Issue #25, opened 2016-09-21
Should we keep the Inject/Extract extension methods for formats?


cwe1ss:
Analogous to known tags ( #18 ), we currently also have `ITracer` extension methods for the built-in formats.

Examples:
```csharp
public static void InjectTextMap(this ITracer tracer, ISpanContext spanContext, IDictionary<string, string> data)
{
    tracer.Inject(spanContext, Formats.TextMap, new DictionaryCarrier(data));
}

public static ISpanContext ExtractTextMap(this ITracer tracer, IDictionary<string, string> data)
{
    return tracer.Extract(Formats.TextMap, new DictionaryCarrier(data));
}

// usage

var data = new Dictionary<string, string>();

tracer.InjectTextMap(existingSpanContext, data);

var extractedSpanContext = tracer.ExtractTextMap(data);
```

The main advantages are:
* Users can pass the actual object instead of having to manually wrap it in a carrier
  * This means, they don't need to know/search for the wrapper types
* People who write their own formats/carrier types can also create their own extension methods so they will show up in the same IntelliSense tooltip.

Are you ok with this or do you see any concerns?

/cc @bensigelman @dawallin @yurishkuro

yurishkuro:
I don't have a strong opinion on these. I do have misgivings in general about overloading the lib with helper functions, and from the past experience with C# I remember that extension methods are a bit hard to follow when reading the code as they are often imported inconspicuously.

bensigelman:
I can see advantages here, but my personal opinion is that any convenience functions should be kept out of the OT layer until there's a clear pattern/problem emerging in actual production calling code. My two cents.

cwe1ss:
Extension methods have been removed with #26 - thank you for your feedback!

PR #26, opened 2016-10-15
Removes all Inject/Extract helpers and extension methods

cwe1ss:
Discussion: #25 

This PR would remove almost all helpers and extension methods for inject/extract. The only helper that's left is the `DictionaryCarrier` type which wraps a `IDictionary<string, string>`.

With this, the API probably has the smallest possible surface. We might reconsider later whether these things should come back into the main API or be moved into a contrib-library.

/cc @bensigelman @dawallin @yurishkuro


cwe1ss:
Even though there hasn't been any feedback here, I'll merge this anyway. The general topic was discussed in #25 and this PR just removes helper code. It also gets the code closer to Java, so there shouldn't be any issues/concern. 

We can bring this back later, should we want to.

PR #28, opened 2016-10-16
Restrict tag types to bool/double/string like in Java

cwe1ss:
I'm writing my first instrumentation for this library and I think, now that we no longer have extension methods on tags, it's way too easy to pass wrong types to `SetTag` and `WithTag`. Examples are passing enums instead of int, Uri instead of string, ...

This PR removes the string:object version and adds 3 overloads for `bool`/`string`/`double`. This is very similar to what Java has except that there isn't a base type for all numeric types in .NET. 

Using `double` will accept everything except `decimal`. If we would add an additional overload for `decimal`, then we have some types (e.g. `integer`) that would cast to both and the compiler will require an explicit cast. For this reason, I think it's ok to leave out decimal for now. If we need it, we probably need even more specific overloads to address the compiler errors.

I'll merge this in 3-4 days if there are no comments by then.

### DIFF
--- a/src/OpenTracing/ISpan.cs
+++ b/src/OpenTracing/ISpan.cs
@@ -24,13 +24,25 @@ namespace OpenTracing
         /// Adds a tag to the Span.
         /// </summary>
         /// <param name="key">If there is a pre-existing tag set for <paramref name="key"/>, it is overwritten.</param>
-        /// <param name="value">
-        /// Tag values can be numeric types, strings, or bools. The behavior of other tag value types is undefined 
-        /// at the OpenTracing level. If a tracing system does not know how to handle a particular value type, it
-        /// may ignore the tag, but shall not panic.
-        /// </param>
+        /// <param name="value">The value to be stored.</param>
         /// <returns>The current <see cref="ISpan"/> instance for chaining.</returns>
-        ISpan SetTag(string key, object value);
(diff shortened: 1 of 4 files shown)


dawallin:
What is the use case for using different types on Tags? Is there any reason not to treat all tags as strings? Is SetTag("TagName", 123) ever treated differently from SetTag("TagName", "123")? If everything is treated as strings I think doing .ToString() on object would be sufficient. 
If we like to keep the code a port from the Java implementation, then I think this PR is the best alternative.

bensigelman:
@dawallin the use case for different types is simply that it encodes more information about the data for downstream tracing systems. E.g., it's possible to tag all outbound RPCs with their size in bytes as an `int` tag, then build a histogram over those byte size values in a monitoring system. Also, using the "real" type allows formatting to be a late-bound decision in UIs (e.g., to use scientific notation or not) rather than something the clients need to take care of themselves.

Not saying it's critically important, but maybe that provides some motivation.

I do think it makes sense for the OT layer to distinguish between types in some way, though I can certainly imagine implementations that choose to coerce everything into a string for whatever reason.

cwe1ss:
ok - seems like we can do this. I'll merge this, since I really want to do a release within the next few days.

We could go to a more generic solution later without breaking anything anyway.

PR #34, opened 2017-04-26
Adds constants for standard log fields

cwe1ss:
This would add constants for the standard log fields from [semantic_conventions.md](https://github.com/opentracing/specification/blob/master/semantic_conventions.md)

A similar thing does not yet exist in Java though.

Also, I'm not sure if I should keep the `stack` field since it's probably never used by anyone. Instrumentation code will always just pass an exception object in `error.object` and the tracer will deal with it.

### DIFF
--- a/src/OpenTracing/LogFields.cs
+++ b/src/OpenTracing/LogFields.cs
@@ -1,0 +1,43 @@
+﻿using System;
+
+namespace OpenTracing
+{
+    /// <summary>
+    /// <para>Every Span log has a specific timestamp and one or more fields.
+    /// This class defines the standard fields.</para>
+    /// <para>See https://github.com/opentracing/specification/blob/master/semantic_conventions.md for details.</para>
+    /// </summary>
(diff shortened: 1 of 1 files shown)


src/OpenTracing/LogFields.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// A concise, human-readable, one-line message explaining the event.
+        /// E.g., "Could not connect to backend", "Cache invalidation succeeded".
+        /// </summary>
+        public const string Message = "message";
+
+        /// <summary>
+        /// A stack trace in platform-conventional format; may or may not pertain to an error.
+        /// E.g., "File \"example.py\", line 7, in \&lt;module\&gt;\ncaller()\nFile \"example.py\", line 5,
```
bhs: probably ought to make this example about C# (in terms of the "platform-conventional format", I mean).

Issue #35, opened 2017-04-26
In-process propagation - Logic for activating/deactivating spans


cwe1ss:
References:
* General discussion: https://github.com/opentracing/specification/issues/23
* Java PR: https://github.com/opentracing/opentracing-java/pull/115
* [.NET DiagnosticSource/Activity](https://github.com/dotnet/corefx/tree/master/src/System.Diagnostics.DiagnosticSource/src/System/Diagnostics)

C# has several features that will likely/hopefully make in-process propagation easier than in Java:
* `async`/`await` does not require the use of promises/futures.
* `AsyncLocal` is a storage that automatically flows with async execution.

The upcoming `Activity` type from .NET (very similar to an OpenTracing `Span`) uses a simple static property `Activity.Current` that uses `AsyncLocal` (.NET 4.6 / .NET Core) or `CallContext` (.NET 4.5) and therefore automatically flows with the async execution. Every `Activity` that is started sets an existing `Activity.Current` as its parent (unless the user manually set a `ParentId`).

However, if we compare this with the current Java PR, the following features are not possible with this system right now:
* It's not possible to start an activity without making it active
* It's not possible to ignore the active parent

What I'd like to discuss in this issue is whether or not these two things are even necessary in C#!

#### Question 1: What non-performance use cases exist, where you would NOT want to make a span active?

Thanks to `async`/`await` simple use cases like instrumenting outgoing async calls just work and wouldn't even need a separate in-process propagation system:

```csharp
public async Task<Response> Send(Request r)
{
    using (ISpan span = _tracer.BuildSpan("someWork").Start())
    {
        Response r = await _client.SendAsync(r);

        span.SetTag("status", r.Status);
    }
}
```
However, this would be problematic in case of the integrated `HttpClient`, because it uses a form of decorator pattern to chain together user logic. If the instrumentation code is not the last in the chain, code from inner decorators is not able to access this span.

Not making the span active is therefore not without issues. Also, given that the instrumented code usually does a network call or similar, the potential overhead of writing the AsyncLocal (I don't have numbers for this yet) might be negligible. 

*From a performance point of view it might therefore very well be ok to always make a span active.*

#### Question 2: What use-cases are there where you would want to ignore the currently active span?

One scenario I can think of right now is a messaging system where you receive an event which was part of a trace and you want to deliberately start a new trace with the work that follows from that event.

Is this a valid scenario for this use case? Are there other examples?

[...]

PR #36, opened 2017-04-27
[WIP] In-process propagation with a simple `ITracer.ActiveSpan` property

cwe1ss:
Related to #35. This PR shows a very simple contract for in-process propagation in C#.

It has the following properties:
* `ITracer.ActiveSpan` holds the currently active span. 
* Every new span will automatically get the active span as a parent.
* Opting out of this inheritance behavior is possible through `ISpanBuilder.IgnoreActiveSpan()`.
* There is no separate `IActiveSpan` because C# doesn't need the "Continuation" feature from the Java PR.
* This is a breaking change because the current usage would automatically activate spans. I don't think this is a problem though because I'm not even sure if anyone is using this library already :smile:

Since this library currently only holds contracts, these rules must be met by tracer implementations. We should discuss separately if there are any building blocks that should go into this library or a contrib project.

Things that are not possible with this right now:
* "Activating" an existing span that has been started with `ISpanBuilder.IgnoreActiveSpan()`
  * It should be discussed in #35 whether this is necessary.

Questions:
* Does a user need an API to access a span's parent? (e.g. `ISpan.Parent` or `ISpanContext.Parent`)

Usage:

```csharp
// Accessing the currently active span.
ISpan activeSpan = tracer.ActiveSpan;

// Starting a new span will automatically inherit an existing active span
// with a "child_of" reference and will "activate" itself.
ISpan span = tracer.BuildSpan("someWork").Start();
try
{
    HttpResponseMessage response = await httpClient.SendAsync(request);

    // C# automatically restores local variables in an async execution,
    // so this line doesn't even need to access `tracer.ActiveSpan`.
    span.SetTag(Tags.HttpStatusCode, response.StatusCode);
}
finally
{
    // Finish will revert "tracer.ActiveSpan" to its parent.
    span.Finish();
}

// The same can be achieved by using the IDisposable feature of ISpan and a "using" statement:
using (ISpan activeSpan = tracer.BuildSpan("someWork").Start())
{
    // (record data to `span`, call async methods, ...)
}

// If an `ActiveSpan` should NOT act as a parent for a newly started span
// a user can invoke `ISpanBuilder.IgnoreActiveSpan()` at `BuildSpan()` time, like so:
ISpan activeSpan = tracer.BuildSpan("someWork")
    .IgnoreActiveSpan()
    .Start();

```

### DIFF
--- a/src/OpenTracing/ISpanBuilder.cs
+++ b/src/OpenTracing/ISpanBuilder.cs
@@ -37,6 +37,11 @@ namespace OpenTracing
         ISpanBuilder AddReference(string referenceType, ISpanContext referencedContext);
 
         /// <summary>
+        /// When used, the newly created span will NOT inherit <see cref="ITracer.ActiveSpan"/> as a parent.
+        /// </summary>
+        ISpanBuilder IgnoreActiveSpan();
+
+        /// <summary>
         /// Adds a tag to the span.
         /// </summary>
         /// <param name="key">If there is a pre-existing tag set for <paramref name="key"/>, it is overwritten.</param>
--- a/src/OpenTracing/ITracer.cs
+++ b/src/OpenTracing/ITracer.cs
@@ -8,6 +8,13 @@ namespace OpenTracing
     public interface ITracer
     {
         /// <summary>
+        /// <para>Returns the currently active span. It is used for intra-process propagation and follows the async execution flow.</para>
+        /// <para>Any newly started span will inherit this span as a parent unless <see cref="ISpanBuilder.IgnoreActiveSpan"/>
+        /// is called before the span is started.</para>
+        /// </summary>
+        ISpan ActiveSpan { get; }
+
+        /// <summary>
         /// Returns a new <see cref="ISpanBuilder" /> for a span with the given <paramref name="operationName" />.
         /// </summary>
         /// <param name="operationName">The operation name of the span.</param>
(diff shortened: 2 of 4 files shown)


tedsuo:
@carlosalberto any thoughts?

carlosalberto:
Definitely sounds all fine to me, specially the part about not needing a few things since async/await help us here (together with AsyncLocal).

For the HttpClient and similar components, it looks like there's not much of an option, other than be explicit about it with the users.

bhs:
cc += @terribledev

TerribleDev:
what is the plan for getting the span into the httpclient request? have you thought at all about making extension methods of HttpClient that will add the ISpan to the request? httpClient.SentAsync(request, span), or a class that inherits from HttpClient but takes an ISpan in the constructor

bmermet:
@cwe1ss Any progress on this? The Java propagation API seems stabilized and will soon be released, I can help porting it to c# if it's the right thing to do.

cwe1ss:
See #52 as well. 

(.NET‘s Activity has built in support for propagation via Activity.Current)

carlosalberto:
An interesting take would be to prototype out an API/wrapper for `Activity` to work under the OpenTracing API. Wondering if that would be useful (for a start), though. Ideas?

cwe1ss:
Out of date. I'll create a new PR with the new scope stuff from Java.

PR #39, opened 2017-05-31
Adds a (very) simple bash-based build-script

### LINKED ISSUE #38: Consider changing the builds?
Right now the builds are done in powershell with psake, which I cannot run on my ubuntu laptop. Do you think it would make more sense to switch the task system to something node, ruby, or even just c#.

I wrote a [gulp plugin](https://www.npmjs.com/package/gulp-dotnet-cli) for the new cli, mostly because my team codes and runs dotnet on *nix.

cwe1ss:
Fixes #38 

This is just a very simple bash build script. Note that it does NOT install the dotnet CLI. it must be installed according to https://www.microsoft.com/net/core

@TerribleDev Will this work for you?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
       dist: trusty
       sudo: required
       mono: none
-      dotnet: 1.0.3
+      dotnet: 1.0.4
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


TerribleDev:
yeah thats fine.

PR #46, opened 2017-10-20
WIP: reorganize usings; refactoring;

monica85rodrigues:
Hi, 
It's a pleasure to see good code :D 
I only reorganize your usings, put inside namespaces.
I'm participating on GIT Octoberfest

Thank you

### DIFF
--- a/src/OpenTracing/ISpan.cs
+++ b/src/OpenTracing/ISpan.cs
@@ -1,8 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace OpenTracing
+﻿namespace OpenTracing
 {
+    using System;
+    using System.Collections.Generic;
+
     /// <summary>
     /// Represents an in-flight span in the opentracing system.
     /// Spans are created by the <see cref="ITracer"/> interface.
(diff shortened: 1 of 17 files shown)


src/OpenTracing/ISpan.cs
```diff
@@ -1,8 +1,8 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace OpenTracing
+﻿namespace OpenTracing
 {
+    using System;
+    using System.Collections.Generic;
```
TerribleDev: Why would one place the usings inside of the namespace block? Also isn't this a change to this repo's existing style and formatting....

monica85rodrigues: Hi, You can see here a good explanation:
https://stackoverflow.com/questions/125319/should-using-statements-be-inside-or-outside-the-namespace

But you know, you can do that way. It's an opinion ;) Thanks.

cwe1ss:
Hi, thanks for taking the time to create this PR. However, I‘m not going to accept it as this is mostly a personal preference thing. It also seems that newer code from Microsoft (eg roslyn, asp.net core) have the usings outside of the namespace.

PR #49, opened 2017-11-07
Inversion of control/separation of SpanBuilder syntax from implementation of ITracer

ndrwrbgs:
The specification suggests a format similar to a SpanBuilder, but does not require such for implementations. Given that, perhaps it would be better to use a StartSpan (as the specification does detail) with it's optional parameters, and make SpanBuilder a thing that operates ON an ITrace. This would prevent every implementor from having to have quite similar implementations, and from a naïve implementor from accidentally emitting tags in the span builder before the span has started (it prevents implementors from having to deal with partially initialized objects).

Though, perhaps we do not like optional parameters, in which case there would be at least 3^2 overloads required, which would nearly mandate that we expose a builder directly from ITracer.

### DIFF
--- a/src/OpenTracing/ITracer.cs
+++ b/src/OpenTracing/ITracer.cs
@@ -1,4 +1,6 @@
 ﻿using OpenTracing.Propagation;
+using System;
+using System.Collections.Generic;
 
 namespace OpenTracing
 {
@@ -8,10 +10,17 @@ namespace OpenTracing
     public interface ITracer
     {
         /// <summary>
-        /// Returns a new <see cref="ISpanBuilder" /> for a span with the given <paramref name="operationName" />.
+        /// Starts a new Span
         /// </summary>
         /// <param name="operationName">The operation name of the span.</param>
-        ISpanBuilder BuildSpan(string operationName);
+        ISpan StartSpan(
+            string operationName,
+            IEnumerable<Tuple<string, ISpanContext>> references = null,
+            DateTimeOffset? explicitStartTime = null,
+            IEnumerable<KeyValuePair<string, string>> stringTags = null,
+            IEnumerable<KeyValuePair<string, bool>> boolTags = null,
+            IEnumerable<KeyValuePair<string, double>> doubleTags = null,
+            IEnumerable<KeyValuePair<string, int>> intTags = null);
 
         /// <summary>
         /// Inject a <see cref="ISpanContext"/> into a <paramref name="carrier"/> of a given type,
--- a/src/OpenTracing/TracerSpanBuilderExtensions.cs
+++ b/src/OpenTracing/TracerSpanBuilderExtensions.cs
@@ -3,6 +3,8 @@ using System.Collections.Generic;
 
 namespace OpenTracing
 {
+    using System.Linq;
+
     public static class TracerSpanBuilderExtensions
     {
         public static ISpanBuilder BuildSpan(ITracer tracer, string operationName)
@@ -17,20 +19,14 @@ namespace OpenTracing
             private string operationName;
             private ICollection<Tuple<string, ISpanContext>> references;
             private DateTimeOffset? explicitStartTime;
-            private ICollection<KeyValuePair<string, string>> stringTags;
-            private ICollection<KeyValuePair<string, bool>> boolTags;
-            private ICollection<KeyValuePair<string, double>> doubleTags;
-            private ICollection<KeyValuePair<string, int>> intTags;
+            private IDictionary<string, object> tags;
 
             public SpanBuilder(ITracer tracer, string operationName)
             {
                 this.tracer = tracer;
                 this.operationName = operationName;
                 this.references = new List<Tuple<string, ISpanContext>>();
-                this.stringTags = new List<KeyValuePair<string, string>>();
-                this.boolTags = new List<KeyValuePair<string, bool>>();
-                this.doubleTags = new List<KeyValuePair<string, double>>();
-                this.intTags = new List<KeyValuePair<string, int>>();
+                this.tags = new Dictionary<string, object>();
             }
 
             public ISpanBuilder AsChildOf(ISpan parent)
@@ -61,25 +57,25 @@ namespace OpenTracing
 
             public ISpanBuilder WithTag(string key, bool value)
             {
-                this.boolTags.Add(new KeyValuePair<string, bool>(key, value));
+                this.tags.Add(key, value);
                 return this;
             }
 
             public ISpanBuilder WithTag(string key, double value)
             {
-                this.doubleTags.Add(new KeyValuePair<string, double>(key, value));
+                this.tags.Add(key, value);
                 return this;
             }
 
             public ISpanBuilder WithTag(string key, int value)
             {
-                this.intTags.Add(new KeyValuePair<string, int>(key, value));
+                this.tags.Add(key, value);
                 return this;
             }
 
             public ISpanBuilder WithTag(string key, string value)
             {
-                this.stringTags.Add(new KeyValuePair<string, string>(key, value));
+                this.tags.Add(key, value);
                 return this;
             }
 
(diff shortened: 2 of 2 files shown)


src/OpenTracing/ITracer.cs
```diff
@@ -8,10 +10,17 @@ namespace OpenTracing
...
-        /// Returns a new <see cref="ISpanBuilder" /> for a span with the given <paramref name="operationName" />.
+        /// Starts a new Span
         /// </summary>
         /// <param name="operationName">The operation name of the span.</param>
-        ISpanBuilder BuildSpan(string operationName);
+        ISpan StartSpan(
+            string operationName,
+            IEnumerable<Tuple<string, ISpanContext>> references = null,
+            DateTimeOffset? explicitStartTime = null,
+            IEnumerable<KeyValuePair<string, string>> stringTags = null,
```
ndrwrbgs: These could be {string, object} pairs, which would remove the requirement to have many arguments for the same thing, but that would require boxing of the value types (is this a blocker though?).

Even if it is, having 4 parameters to mean the same thing is bad, and we should perhaps (if accepting this change) change to use some kind of Tag struct that can have string OR bool OR double OR int (would require either a large struct [bad] or a class with an interface - which would require boxing same as object, but be more type safe).

src/OpenTracing/ITracer.cs
```diff
@@ -8,10 +10,17 @@ namespace OpenTracing
...
     {
         /// <summary>
-        /// Returns a new <see cref="ISpanBuilder" /> for a span with the given <paramref name="operationName" />.
+        /// Starts a new Span
         /// </summary>
         /// <param name="operationName">The operation name of the span.</param>
-        ISpanBuilder BuildSpan(string operationName);
+        ISpan StartSpan(
+            string operationName,
+            IEnumerable<Tuple<string, ISpanContext>> references = null,
```
ndrwrbgs: This should have a type, but Reference seemed too close to References. If we accept this, should we rename References to KnownReferences, or make a Reference type and have the constants exposed as static members with that type (seems like violation of SRP).

src/OpenTracing/TracerSpanBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+        private sealed class SpanBuilder : ISpanBuilder
+        {
+            private readonly ITracer tracer;
+
+            private string operationName;
+            private ICollection<Tuple<string, ISpanContext>> references;
+            private DateTimeOffset? explicitStartTime;
+            private IDictionary<string, object> tags;
```
ndrwrbgs: The second commit in this PR introduced this unified dictionary in place of individual collections. This is likely less CPU performant due to boxing, but makes for more clear code.

An alternative but relatively equal change would be to keep the separate collections and a set of known keys (to ensure duplicate keys with different values are handled properly).

src/OpenTracing/TracerSpanBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+
+            public SpanBuilder(ITracer tracer, string operationName)
+            {
+                this.tracer = tracer;
+                this.operationName = operationName;
+                this.references = new List<Tuple<string, ISpanContext>>();
+                this.tags = new Dictionary<string, object>();
```
ndrwrbgs: Given these are frequently used in relatively hot paths, should not be doing allocations unless required (lazy init the dictionary?)

src/OpenTracing/TracerSpanBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+            private string operationName;
+            private ICollection<Tuple<string, ISpanContext>> references;
+            private DateTimeOffset? explicitStartTime;
+            private IDictionary<string, object> tags;
+
+            public SpanBuilder(ITracer tracer, string operationName)
+            {
+                this.tracer = tracer;
+                this.operationName = operationName;
+                this.references = new List<Tuple<string, ISpanContext>>();
```
ndrwrbgs: Initial sizes for the list and dictionary should be supported for optimization purposes.

src/OpenTracing/TracerSpanBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+        private sealed class SpanBuilder : ISpanBuilder
+        {
+            private readonly ITracer tracer;
+
+            private string operationName;
+            private ICollection<Tuple<string, ISpanContext>> references;
+            private DateTimeOffset? explicitStartTime;
+            private IDictionary<string, object> tags;
+
+            public SpanBuilder(ITracer tracer, string operationName)
```
ndrwrbgs: Some implementations may forgo initialization logic if tracing on the parent span is disabled (to enable hot path code).

Three obvious solutions to this:
1. Users can implement their own alternative extension methods
2. Users can implement ALL possible extension methods (just remove ISpanBuilder entirely from this interfaces project
3. Leave as is, requiring users to implement ISpanBuilder

src/OpenTracing/TracerSpanBuilderExtensions.cs
```diff
@@ -0,0 +1,101 @@
...
+                return this;
+            }
+
+            public ISpan Start()
+            {
+                return this.tracer.StartSpan(
+                    this.operationName,
+                    this.references,
+                    this.explicitStartTime,
+                    this.tags.Where(kvp => kvp.Value is string).Select(kvp => new KeyValuePair<string, string>(kvp.Key, (string) kvp.Value)),
```
TerribleDev: if you change the following 4 linq statements to a foreach statement you can save 3 iterations over the same array.

TerribleDev:
Personally I don't mind optional args

src/OpenTracing/TracerSpanBuilderExtensions.cs
ndrwrbgs: LINQ doesn't iterate over the collection for each statement. This will result in one iteration, same as nested foreaches

ndrwrbgs:
== Optional Parameters == RE @TerribleDev 
I don't either :smile: I find them super useful. Though it is a decision we shouldn't take lightly for a library, due to how .NET handles them. The main concerns would be
* It seems the language spec may have some gotcha's on optional parameters, leading to CA1026's inception.[[link]](https://msdn.microsoft.com/en-us/library/ms182135.aspx)
* There are some versioning gotcha's on changes to methods with optional parameters/introductions of overloads[[link]](https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/) (I think this behavior even caused some problems adopting .net 4.6[[link]](https://social.msdn.microsoft.com/Forums/en-US/5ad7cf7e-6a54-4768-82d3-e55cd46d56b5/method-not-found-systemstringformat?forum=csharpgeneral)
* I've also heard that they're compiled in (across assembly boundaries) which means you can't just link to a new dll when a default changes, you have to recompile the consumer to get the new default - but I'm not 100% sure on that first-hand

So while I'm all for having them, we should be sure that we're okay with the caveats (and able to assist folks if they come asking when they hit these things).

TerribleDev:
@ndrwrbgs  I thought all of these were well understood. Ie, if you add new parameters that are optional you are changing the method signature and thus its a breaking change. Just because the source still compiles doesn't mean you havn't introduced breakages.

TerribleDev:
Here is a good article about them from way back: https://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

ndrwrbgs:
(You realize you just linked me to what I linked you to, right? :smiley:)

cwe1ss:
This is quite a big move away from the Java version and the SpanBuilder implementation makes quite a lot of assumptions which I think we shouldn't make until we better understand how this library will be used in the .NET world.

So I'd vote for not doing this right now.

ndrwrbgs:
@cwe1ss I'd agree with this for now. Does java have extension methods? If so I"d suggest the change over there, as these methods have no business being required to implement for implementors given their implementation is strongly specified as far as what they'll do (making them prime candidates for an extension method).

TL;DR java should do this first and us second, to negate the possibility of java making changes that change the model of these methods.
Follow up item: does java have extension methods and how would we request that change from someone?

cwe1ss:
AFAIK, Java does not have extension methods. If you want to have this discussed you can either create an issue in https://github.com/opentracing/specification (if it's valuable for all languages) or directly in https://github.com/opentracing/opentracing-java (if you want to discuss it with the java people).

I'll close this for now. Anyone feel free to open a new issue if you want to have it re-opened.

Issue #47, opened 2017-11-07
Parity with the other libraries


ndrwrbgs:
The opentracing-csharp library seems to not get much love & attention lately. I was curious if there was interest in bringing it up to parity with more active/maintained opentracing-java, since the languages are similar enough that updating to or even porting the code would be trivial. Many concepts one may want are already there (such as those mentioned directly or indirectly by #36, #44, #35).

This question is obviously motivated by the assumption that given that the java implementation has a large user base, changes to it's exposed API may end up being more _practically speaking_ official than changes actually in the specifications. If, alternatively, it's desirable to keep this library _exactly_ matching the specs with nothing more, then that should be decided/noted explicitly, which would give guidance to the proper action with PRs like #36.

cwe1ss:
Sorry for not replying - I’m on a sabbatical right now. Please see https://github.com/opentracing/opentracing-csharp/issues/52 for a topic that IMO should be resolved before we do any further changes.

cwe1ss:
I'm back and I'll have at least a few days where I can work on OpenTracing. If you guys are ok with it, I'll create several PRs to bring this library up to date with the Java version. I'd rather not merge #50 as it is right now because it doesn't follow C# guidelines and it changes some things which I think should not be changed so I'd like to at least have them discussed briefly. But as I've said there, I'd be happy to wait for changes/other PRs/feedback from @ndrwrbgs as I haven't been active here so I don't want to make anyone unhappy :)

/cc @tedsuo @carlosalberto (I can't see if anyone else is a owner/contributor)

cwe1ss:
No more need for this issue - the work to ensure parity is ongoing in #55 and several PRs.

Issue #44, opened 2017-10-15
Use IReadonlyDictionary<TKey, TValue> instead of IEnumerable<KeyValuePair<TKey, TValue>>


gideonkorir:
I'm proposing that we change methods that use ```IEnumerable<KeyValuePair<TKey,TValue>>``` to ```IReadonlyDictionary<TKey,TValue>```. I believe that using ```IReadonlyDictionary``` maintains the semantics  of an immutable map but is easier to work with e.g. we can use indexer ```map[key]``` and also access ```Count``` property.

If agreed we can make the following changes:

```
interface ISpanContext {
 IReadonlyDictionary<string, string> BaggageItems { get; }
}

interface ISpan
{
  Log(IReadonlyDictionary<string, string> fields);
  Log(DateTimeOffset timestamp, IReadonlyDictinary<string, string> fields)
}

````

Using ```IReadonlyDictionary<TKey,TValue>``` simplifies implementations e.g. 

1. Duplicate key behavior is undefined for ```Log(IEnumerable)``` so using ```IReadOnlyDictionary``` averts that behavior
2. In the ```Log(IEnumerable)``` it's simpler to check for semantic fields e.g. Brave library checks for ```event``` key

gideonkorir:
Maybe we also change ```string ISpan.GetBaggageItem(string key)``` to ```bool ISpan.TryGetBaggageItem(string key, out string value)```. This makes explicit that the key could be missing and doesn't require the client to check for nulls.

ndrwrbgs:
As duplicate keys are by definition undefined, it seems to make sense to make the contract represent that fact (ICollection<KeyValuePair> is slightly semantically different from IReadOnlyDictionary<KeyValuePair>).

[Note c#’s IReadOnlyDictionary isn’t actually enforced to be immutable, for those coming from Java it may be more clear to think of it as IReadableDictionary — and interface shouldn’t say what an implementor CAN’T do, but that’s for another time].

This change would be more in line with what the java implementation does, which is expose a map).

The one con of this change is that it adds an implementation constraint that is not present in the specification. The specification does bring forth the question of duplicate keys like this would address, but it does not require that baggage items, on transfer, be materialized (e.g. imagine passing around the baggage serialized, and being able to streaming-wise deserialize it. Based PURELY on the specification, an implementor would be capable of making every call to BaggageItems make a database call, but if we expose IReadOnlyDictionary a caller would expect that to NOT be the case - [index] calls should be quick).

[...]

Issue #42, opened 2017-07-21
Replace DateTime usage with DateTimeOffset


hotchkj:
Really pleased to see OpenTracing start up for .NET! Based on what they're used for, I wanted to suggest that the various interfaces be updated to use `DateTimeOffset` for logging etc. This would be really helpful for implementers (for example the company I work for is implementing a tracer for use internally).

This would match [MSDN docs](https://blogs.msdn.microsoft.com/davidrickard/2012/04/06/datetime-and-datetimeoffset-in-net-good-practices-and-common-pitfalls/) on the subject. There's also a great [StackOverflow](https://stackoverflow.com/questions/4331189/datetime-vs-datetimeoffset) topic providing an analogy for why `DateTimeOffset` should be used for this sort of tracing.

Top of the list is that `DateTimeOffset` serialises unambiguously which I believe will be a big help in implementing propagators correctly.

Example of Serilog logging framework using this for timestamps: https://github.com/serilog/serilog/blob/dev/src/Serilog/Events/LogEvent.cs#L52

cwe1ss:
See https://github.com/opentracing/opentracing-csharp/issues/4 for the reasons why we chose `DateTime`.

Please post your scenario if you disagree.

hotchkj:
The intent of the interface `ISpan Log(DateTime timestamp, string eventName);` appears to be to log an occurrence at a specific point in time. Providing a `DateTimeOffset` is always correct & unambiguous, and in codebases that are following Microsoft guidance, all of the existing code would already be using `DateTimeOffset` for tracking timestamps (as I indicated, major .NET logging frameworks already do).

The choice of `DateTime` in the interface means we must convert from `DateTimeOffset` everywhere to `DateTime` which returns, as per docs, `The value of the DateTime.Kind property of the returned DateTime object is DateTimeKind.Unspecified.` which gives exactly the problem you outlined in #4 - every time we convert to OpenTracing format, we have to be aware of all of the pitfalls that `DateTimeOffset` is designed to avoid. This makes the library harder to work with, not easier.

I can provide more arguments, but I'd stand by the original linked article - Microsoft have made it very clear that modern .NET code should prefer `DateTimeOffset` wherever possible.

cwe1ss:
DateTimeOffset is only useful when you want to store the offset (e.g. for local user dates). This (IMO) does not make much sense for server scenarios where you always need UTC.

[...]

PR #50, opened 2017-11-08
Migrate/copy opentracing-java's content

ndrwrbgs:
Performing #47 to have it available if it is desired


ndrwrbgs:
Closing PR to reduce noise - this PR is for demonstrative purposes and not actually intended for review unless the corresponding discussion page is resolved.

ndrwrbgs:
I've diffed the contents of the two, and the primary differences are around the introduction of the `ActiveSpan` related code - the rest are comment based and superficial - they may be useful in being latest (as it seems this package was originally created off the java one anyway), and could be factored out as desired

ndrwrbgs:
Whoops, they were off in a branch developing new awesomeness! I ported their old stuff 😞 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

ndrwrbgs:
Reopening now that it's January, when the investment in other languages was scheduled to start. Will use this branch for porting the "new" stuff also.

Aaronontheweb:
@ndrwrbgs looks like the build is failing due to a missing .NET CLI binary, right?

cwe1ss:
Thanks for taking the time to create this PR and sorry for not responding earlier - I've been off the grid for quite some time! 

This PR goes a bit too far IMO as it is now a "blind" copy of Java without taking C# style guides and constructs into consideration. 
It e.g. uses static readonly instead of constants, it no longer uses properties in some locations, it doesn't follow .NET naming guidelines (e.g. "BaggeItems()" instead of "GetBaggageItems()"), it changes the programming style of the whole repository (e.g. usings within namespaces - which I'd prefer not to do as this is not what Microsoft does in their recent repos like ASP.NET Core etc) and it also changes functionality (e.g. IDictionary instead of IEnumerable for logs). It also uses wrong license headers as this repo is currently on a MIT license.

I'd prefer to have functional changes (like IEnumerable->IDictionary, java-style tags, ...) in separate issues/PRs to give people a chance to comment on them. It's also a lot easier to review stuff as this is one MEGA PR. 

Do you want to do some changes here and do you have time to do so within the next few days? I'm asking because I have some free time now and I want to get as much stuff done in this time as possible. But I obviously also don't want to have your efforts unnoticed in the commit history.

ndrwrbgs:
I’m fine iterating and sounds like it’d be less work than you reinventing the library. My concern is that you want to avoid functional changes, but the new APIs are littered with them (on the java side), and you’re asking for some deviations from the java one. Both of these are fine to me but I’d personally presume deviating from the “common patterns across the la gauge agnostic standard” would be things that should be individually reviewed, rather than the adoption of the standard. Thoughts?

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="AbstractTag.cs">
```
cwe1ss: We currently use MIT which is also used by .NET Core so I guess it should be fine. I don't care which license we use but if we want to change it, we need approval from someone. So I'd rather stick to MIT for now since that's what's live on nuget.org 

I'd also prefer not to have license headers in files at all since they are a burden to maintain. Also, implementations like opentracing-go don't use them.

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
...
+//   Unless required by applicable law or agreed to in writing, software distributed under the License
+//   is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+//   or implied. See the License for the specific language governing permissions and limitations under
+//   the License.
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OpenTracing
+{
+    using System;
```
cwe1ss: I want to use coding guidelines that follow recent OSS practices by Microsoft and I think .NET core and ASP.NET Core are the go-to for this. Following this, usings should stay outside of the namespace at the beginning of the file.

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    /// <summary>
+    ///     A <see cref="IScope" /> formalizes the activation and deactivation of a <see cref="ISpan" />, usually from a CPU
+    ///     standpoint.
+    ///     <para>
+    ///         Many times a <see cref="ISpan" /> will be extant (in that <see cref="ISpan.Finish()" /> has not been called)
+    ///         despite being in a
```
cwe1ss: weird line breaks

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
...
+    ///         non-runnable state from a CPU/scheduler standpoint. For instance, a <see cref="ISpan" /> representing the child
+    ///         side of an
+    ///         RPC will be unfinished but blocked on IO while the RPC is still outstanding. A <see cref="IScope" /> defines
+    ///         when a given
+    ///         <see cref="ISpan" /> <em>is</em> scheduled and on the path.
+    ///     </para>
+    ///     <para>
+    ///         Calling <see cref="IScope.Dispose" /> marks the end of the active period for the current thread and
```
cwe1ss: This might not build when XML docs are created (with RELEASE) as Dispose is on IDisposable. Not sure though.

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
...
+    ///         <see cref="IScope" />,
+    ///         updating the <see cref="IScopeManager.Active" /> in the process.
+    ///     </para>
+    /// </summary>
+    public interface IScope : IDisposable
+    {
+        /// <summary>
+        ///     The <see cref="ISpan" /> that's been scoped by this <see cref="IScope" />
```
cwe1ss: Are you using ReSharper or some other custom style conventions? Visual Studio does not intend comments and it doesn't add spaces before a "/>" by default. I'd prefer using the defaults since that's what Microsoft (.NET Core) and most people use.

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// <summary>
+    ///     The <see cref="IScopeManager" /> interface abstracts both the activation of <see cref="ISpan" /> instances (via
+    ///     <see cref="Activate(ISpan)" />) and access to an active <see cref="ISpan" />/<see cref="IScope" />
+    ///     (via <see cref="Active" />.
+    ///     <seealso cref="IScope" />
+    ///     <seealso cref="ITracer.ScopeManager" />
+    /// </summary>
+    public interface IScopeManager
```
cwe1ss: This should be in a separate file

src/OpenTracing/IScope.cs
```diff
@@ -0,0 +1,95 @@
...
+        /// <summary>
+        ///     Make a <see cref="ISpan" /> instance active.
+        /// </summary>
+        /// <param name="span">The <see cref="ISpan" /> that should become the <see cref="Active" /></param>
+        /// <returns>
+        ///     A <see cref="IScope" /> instance to control the end of the active period for the <see cref="ISpan" />.
+        ///     Span will automatically be finished when <see cref="IScope.Dispose" /> is called. It is a
+        ///     programming error to neglect to call <see cref="IScope.Dispose" /> on the returned instance.
+        /// </returns>
+        IScope Activate(ISpan span);
```
cwe1ss: Java doesn't have this overload

src/OpenTracing/ISpan.cs
```diff
@@ -1,164 +1,177 @@
...
         /// <summary>
-        /// Returns the <see cref="ISpanContext"/> that encapsulates span state that should propagate across process boundaries.
-        /// Note that the return value of <see cref="Context"/> is still valid after a call to <see cref="Finish()"/>, as is
-        /// a call to <see cref="Context"/> after a call to <see cref="Finish()"/>.
+        ///     Retrieve the associated SpanContext.
+        ///     This may be called at any time, including after calls to <see cref="ISpan.Finish()" />.
         /// </summary>
-        ISpanContext Context { get; }
+        /// <returns>The SpanContext that encapsulates Span state that sould propagate across process boundaries.</returns>
+        ISpanContext Context();
```
cwe1ss: This should definitely stay a property

src/OpenTracing/ISpan.cs
```diff
@@ -1,164 +1,177 @@
...
-        ///     { "event", "soft error" },
-        ///     { "type", "cache timeout" },
-        ///     { "waited.millis", 1500 }
-        /// });
-        /// </code>
-        /// </example>
-        ISpan Log(IEnumerable<KeyValuePair<string, object>> fields);
+        /// <returns>The Span, for chaining</returns>
+        /// <seealso cref="Log(string)" />
+        ISpan Log(IDictionary<string, object> fields);
```
cwe1ss: Using IDictionary has quite a big performance hit and it no longer allows the usage of linked lists. IEnumerable<KeyValuePair...> allows for all scenarios and is also used in .NET's Activity type. The drawback is that it potentially allows duplicates. I'd like to have this discussed properly in a separate issue - but this could also happen after a merge of course.

src/OpenTracing/ISpan.cs
```diff
@@ -1,164 +1,177 @@
...
         ISpan SetBaggageItem(string key, string value);
 
         /// <summary>
-        /// Returns the value of the baggage item identified by the given <paramref name="key"/>,
-        /// or <c>null</c> if no such item could be found.
+        ///     The value of the baggage item identified by <paramref name="key" />, or null if no such item could be found.
         /// </summary>
-        /// <param name="key">The name of the key which was used to store the baggage item.</param>
-        string GetBaggageItem(string key);
+        string GetBaggaggeItem(string key);
```
cwe1ss: spelling error

src/OpenTracing/ISpan.cs
```diff
@@ -1,164 +1,177 @@
...
+        ///     Sets an explicit end timestamp and records the span.
+        ///     <para>
+        ///         With the exception of calls to Span.context(), this should be the last call made to the span instance, and to
+        ///         do otherwise leads to undefined behavior.
+        ///     </para>
         /// </summary>
-        void Finish(DateTimeOffset finishTimestamp);
+        /// <param name="finishTime">An explicit finish time</param>
+        /// <seealso cref="ISpan.Context" />
+        void Finish(DateTimeOffset finishTime);
```
cwe1ss: everything else uses "timestamp" so this should stay "finishTimestamp" IMO

src/OpenTracing/ISpanBuilder.cs
```diff
@@ -1,82 +1,174 @@
...
-        /// Returns the started span.
+        ///     Like <see cref="StartActive()" />, but the returned <see cref="ISpan" /> has not been registered via the
+        ///     <see cref="IScopeManager" />.
         /// </summary>
+        /// <returns>
+        ///     The newly-started Span instance, which as *not* been automatically registered
+        ///     via the <see cref="IScopeManager" />
+        /// </returns>
+        /// <seealso cref="StartActive()" />
+        ISpan StartManual();
```
cwe1ss: StartManual is deprecated in Java so we don't even need it

src/OpenTracing/ISpanContext.cs
```diff
@@ -1,19 +1,36 @@
...
     {
-        /// <summary>
-        /// GetBaggageItems grants access to an immutable copy of all baggage items
-        /// stored in the <see cref="ISpanContext"/>.
-        /// </summary>
-        IEnumerable<KeyValuePair<string, string>> GetBaggageItems();
+        /// <returns>All zero or more baggage items propagating along with the associated Span</returns>
+        /// <seealso cref="ISpan.SetBaggageItem" />
+        /// <seealso cref="ISpan.GetBaggaggeItem" />
+        IEnumerable<KeyValuePair<string, string>> BaggageItems();
```
cwe1ss: Using Get* is a well-established .NET convention that should be kept.

src/OpenTracing/ITracer.cs
```diff
@@ -1,33 +1,112 @@
...
+        /// <typeparam name="T">The carrier type, which also parametrizes the Format.</typeparam>
+        /// <param name="spanContext">The SpanContext instance to inject into the carrier</param>
+        /// <param name="format">The Fromat of the carrier</param>
+        /// <param name="carrier">
+        ///     The carrier for the SpanContext state. All Tracer.Inject implementations must support
+        ///     <see cref="TextMap" /> and <see cref="Stream" />
+        /// </param>
+        /// <seealso cref="IFormat{T}" />
+        /// <seealso cref="BuiltinFormats" />
+        void Inject<T>(ISpanContext spanContext, IFormat<T> format, T carrier);
```
cwe1ss: It's not common in .NET to use just "T" - why drop a more explicit "TCarrier"?

src/OpenTracing/Log/Fields.cs
```diff
@@ -0,0 +1,57 @@
...
+
+namespace OpenTracing.Log
+{
+    /// <summary>
+    ///     The following log fields are recommended for instrumentors who are trying to capture more
+    ///     information about a logged event. Tracers may expose additional features based on these
+    ///     standardized data points.
+    ///     <see href="https://github.com/opentracing/specification/blob/master/semantic_conventions.md" />
+    /// </summary>
+    public static class Fields
```
cwe1ss: I know this is a deviation from Java but what sense does it make to have a namespace with just one Type? Why not keep using "LogFields" in the root namespace which is more discoverable and a more descriptive name?

src/OpenTracing/Log/Fields.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        /// <summary>
+        ///     The type or "kind" of an error (only for event="error" logs). E.g., "Exception", "OSError"
+        /// </summary>
+        public static readonly string ErrorKind = "error.kind";
```
cwe1ss: I think these things should stay constants as they will never change at runtime and therefore avoid a method call.

src/OpenTracing/Log/Fields.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        /// <summary>
+        ///     A concise, human-readable, one-line message explaining the event. E.g., "Could not connect
+        ///     to backend", "Cache invalidation succeeded"
+        /// </summary>
+        public static readonly string Message = "message";
+
+        /// <summary>
+        ///     A stack trace in platform-conventional format; may or may not pertain to an error.
+        /// </summary>
```
cwe1ss: add the sample stack from the existing code

src/OpenTracing/Propagation/TextMap.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        /// <summary>
+        ///     Puts a key:value pair into the TextMapWriter's backing store.
+        /// </summary>
+        /// <param name="key">A string, possibly with constraints dictated by the particular Format this TextMap is paired with</param>
+        /// <param name="value">A string, possibly with constraints dictated by the particular Format this TextMap is paired with</param>
+        /// <seealso cref="ITracer.Inject{T}" />
+        /// <seealso cref="BuiltinFormats.TextMap" />
+        /// <seealso cref="BuiltinFormats.HttpHeaders" />
+        void Put(string key, string value);
```
cwe1ss: `Put` is a Java term that is not common in .NET collections. .NET usually uses `Set` instead

src/OpenTracing/Propagation/TextMapExtractAdapter.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    ///     A TextMap carrier for use with Tracer.extract() ONLY (it has no mutating methods).
+    ///     Note that the TextMap interface can be made to wrap around arbitrary data types (not just Dictionary{string,
+    ///     string}
+    ///     as illustrated here).
+    /// </summary>
+    /// <seealso cref="ITracer.Extract{T}" />
+    public sealed class TextMapExtractAdapter : TextMap
+    {
+        private readonly IDictionary<string, string> dictionary;
```
cwe1ss: `_dictionary`. See https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md for more details on what I'd like to stick to. We should link this in a contributing-file.

src/OpenTracing/Propagation/TextMapInjectAdapter.cs
```diff
@@ -0,0 +1,55 @@
...
+    /// <summary>
+    ///     A TextMap carrier for use with <see cref="ITracer.Inject{T}" /> ONLY (it has no read methods).
+    ///     Note that the TextMap interface can be made to wrap around arbitrary data types (not just Dictionary{string,
+    ///     string}
+    ///     as illustrated here).
+    /// </summary>
+    /// <seealso cref="ITracer.Inject{T}" />
+    public sealed class TextMapInjectAdapter : TextMap
+    {
+        private readonly IDictionary<string, string> dictionary;
```
cwe1ss: _dictionary - see above

src/OpenTracing/References.cs
```diff
@@ -1,20 +1,35 @@
...
+    ///     <see cref="References" /> is essentially a namespace for the official OpenTracing reference types.
+    ///     References are used by <see cref="ITracer.BuildSpan" /> to describe the relationships between Spans.
     /// </summary>
+    /// <seealso cref="ISpanBuilder.AddReference" />
     public static class References
     {
         /// <summary>
-        /// See https://github.com/opentracing/specification/blob/master/specification.md#references-between-spans
-        /// for more information about CHILD_OF references.
+        ///     See http://opentracing.io/spec/#causal-span-references for more information about CHILD_OF references
```
cwe1ss: these links are broken in the Java version

src/OpenTracing/Tag/AbstractTag.cs
```diff
@@ -0,0 +1,31 @@
...
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OpenTracing.Tag
+{
+    public abstract class AbstractTag<T>
+    {
+        protected AbstractTag(string tagKey)
+        {
+            this.Key = tagKey;
```
cwe1ss: don't use "this."

src/OpenTracing/Tag/Tags.cs
```diff
@@ -0,0 +1,139 @@
...
+    ///     semantic information about the spans. Tracers may expose additional features based on these
+    ///     standardized data points. Tag names follow a general structure of namespacing.
+    ///     <see href="https://github.com/opentracing/specification/blob/master/semantic_conventions.md" />
+    /// </summary>
+    public static class Tags
+    {
+        /// <summary>
+        ///     A static readonlyant for setting the span kind to indicate that it represents a server span.
+        /// </summary>
+        public static readonly string SpanKindServer = "server";
```
cwe1ss: consts

src/OpenTracing/Tag/Tags.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        /// <summary>
+        ///     DbType indicates the type of Database.
+        ///     For any SQL database, "sql". For others, the lower-case database category, e.g. "cassandra", "hbase", or "redis"
+        /// </summary>
+        public static readonly StringTag DbType = new StringTag("db.type");
+
+        /// <summary>
+        ///     DbInstance indicates the instance name of Database.
+        ///     If the jdbc.url="jdbc:mysql://127.0.0.1:3306/customers", instance name is "customers".
```
cwe1ss: there's no jdbc in .NET

src/OpenTracing/Tag/Tags.cs
```diff
@@ -0,0 +1,139 @@
...
+        public static readonly StringTag DbInstance = new StringTag("db.instance");
+
+        /// <summary>
+        ///     DbUser indicates the user name of Database, e.g. "readonly_user" or "reporting_user"
+        /// </summary>
+        public static readonly StringTag DbUser = new StringTag("db.user");
+
+        /// <summary>
+        ///     DbStatement records a database statement for the given database type.
+        ///     For db.type="SQL", "SELECT /// FROM wuser_table". For db.type="redis", "SET mykey "WuValue".
```
cwe1ss: the Select statement seems broken here (///)

src/OpenTracing/UnsupportedFormatException.cs
```diff
@@ -1,21 +1,39 @@
...
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
 
 namespace OpenTracing
 {
+    using System;
+
     /// <summary>
-    /// This exception should be used when the provided format value is unknown or disallowed by the <see cref="ITracer"/>.
+    ///     This exception should be used when the provided format value is unknown or disallowed by the <see cref="ITracer" />
```
cwe1ss: See previous comment. That seems to be your specific coding style. There should be no indentation and no space before "/>".

src/OpenTracing/IScope.cs
ndrwrbgs: Odd, I copied the license from the java files. We release under different licenses?! Sounds like a nightmare :)

Anyway, if you're okay w/o licenses I am definitely too, especially since there's no IP concerns really since it's open source. Will remove

src/OpenTracing/IScope.cs
ndrwrbgs: Silly Microsoft, can't even agree consistently in what to suggest to the public, our internal stuff says the opposite (there's some weird dependency resolving/performance thingie with where they're placed that makes inside better for some cases, but historically people prefer outside).

Not blocking so I'll update, thanks!

src/OpenTracing/IScope.cs
ndrwrbgs: #AutomationMadeMeDoIt

src/OpenTracing/IScope.cs
ndrwrbgs: I believe it should since it's also on `IScope` (via `IDisposable` and lacking explicit implementation (can you do explicit implementation on an interface even?)). How can I verify, build release and check the *.xml files?

src/OpenTracing/IScope.cs
ndrwrbgs: Baaaaah I hate the standards that use "XML"doc and don't use XML indentations, so hard to read! But that's fine everyone seems to agree with you so I'll make the tags line up to the text (sad face)

src/OpenTracing/IScope.cs
ndrwrbgs: I think they got rid of it, I'll update the review to the latest. To confirm, is that what's in master or do you know if they have a staging branch I should work off of now?

src/OpenTracing/ISpan.cs
ndrwrbgs: Is i t generally assumed that the ISpan will cache the value of it's ISpanContext? If the operation will potentially have allocations/processing it's best to use a method since developers are more familiar with expecting such things from a method (whereas a property they expect to work [more or less] like a field and be very quick and not necessary to cache themselves)

src/OpenTracing/ISpan.cs
ndrwrbgs: The concern is that you cannot index into IEnumerable (in fact demoting anything down to IEnumerable is usually a perf hit).

That said, this is the `Log` method, I'm not sure why I changed this one :) Unless there's reasonable assumption that an implementation may want a specific field and not want to iterate over everything. In that case, this is >> perf

RE: no longer allows the usage of linked lists, you can implement IDictionary with a LinkedList (doing so means finding a particular item is inefficient, but no more inefficient than doing so on LinkedList directly). It'd be possible to provide an extension method that can take IEnumerable<KeyValuePair> or, if implementors will really really want indexing into it to be performant then providing two methods, one that takes IEnumerable and one that takes IDictionary so that implementor can appropriately react to the functionality enabled by the input

src/OpenTracing/ISpanBuilder.cs
ndrwrbgs: THANK LORDY IF WE CAN GET RID OF IT! <3 <3
That said, `StartActive` is documented to call this internally in it's implementation, so it would be necessary to keep it, would it not? I will review the latest in the javas

src/OpenTracing/ISpanContext.cs
ndrwrbgs: Agreed
[I think somewhere in this I also use the .Put style from java that c# calls .Add, I need to confirm that, note to self in this comment]

src/OpenTracing/Log/Fields.cs
ndrwrbgs: I normally wouldn't really care about how many types are in a namespace (it seems rather pedantic of a metric). If logically, the item belongs in a different 'folder' if you sorted/grouped things, it deserves its own namespace. I think the 'rule' to keep it high is because many people may overuse namespaces and not use them for grouping.

That said, I don't really want to argue style in OSS, so I will change per your suggestion as it makes little difference to me.

src/OpenTracing/Log/Fields.cs
ndrwrbgs: No method call for a field, an JIT should inline it anyway, but yes, for these that won't change permitting COMPILER inlining like consts do would be good.

src/OpenTracing/Log/Fields.cs
ndrwrbgs: Sorry, is this sample stack in the java code or the spec (`semantics.md`) or somewhere else?

src/OpenTracing/IScope.cs
cwe1ss: fyi: There's an issue regarding different licenses in the specifications-project: https://github.com/opentracing/specification/issues/61

Any discussions about changing the license should happen there and then we'll do whatever the result of that issue is.

src/OpenTracing/IScope.cs
cwe1ss: you only have to call ".\build.ps1" in the project root. this will build with RELEASE and throw an error if the xml docs cannot be created properly.

src/OpenTracing/IScope.cs
cwe1ss: I know... I had my own (different) preferences (tabs, anyone?!? :smile:) for a long time but then I just gave up and now I'm just trying to do whatever the VS defaults do as that is what most people have. :|

src/OpenTracing/IScope.cs
cwe1ss: AFAIK master is the latest as they've just released their stuff.

src/OpenTracing/ISpan.cs
cwe1ss: hmm... I'm not sure TBH but I think it's usually just a reference. I'll look into usages tomorrow.

src/OpenTracing/ISpan.cs
cwe1ss: True, I should have written "Using the concrete `Dictionary`-type has quite a big performance hit" (since it does a uniqueness check on every addition). IDK - we should discuss this in #44

src/OpenTracing/Log/Fields.cs
cwe1ss: I added this to #55 - I can live with whatever most people want. It's not a blocker :)

src/OpenTracing/Log/Fields.cs
cwe1ss: In the existing c# code - we had:
```csharp
        /// <summary>
        /// A stack trace in platform-conventional format; may or may not pertain to an error.
        /// <para />E.g.,
        /// <para />at App.Program.SomeMethod() in C:\app\Program.cs:line 14
        /// <para />at App.Program.Main(String[] args) in C:\app\Program.cs:line 9
        /// </summary>
        public const string Stack = "stack";
```

src/OpenTracing/Tag/Tags.cs
cwe1ss: I've seen that there's also comments with JDBC in the current code so not a big issue - we can change this later.

src/OpenTracing/IScope.cs
ndrwrbgs: resolved

src/OpenTracing/IScope.cs
ndrwrbgs: resolved

src/OpenTracing/IScope.cs
ndrwrbgs: resolved

src/OpenTracing/IScope.cs
ndrwrbgs: resolved

src/OpenTracing/IScope.cs
ndrwrbgs: resolved

src/OpenTracing/IScope.cs
ndrwrbgs: resolved

src/OpenTracing/ISpan.cs
ndrwrbgs: resolved

src/OpenTracing/ISpan.cs
ndrwrbgs: resolved

src/OpenTracing/ISpanContext.cs
ndrwrbgs: resolved

src/OpenTracing/ITracer.cs
ndrwrbgs: resolved

src/OpenTracing/Log/Fields.cs
ndrwrbgs: resolved

src/OpenTracing/Log/Fields.cs
ndrwrbgs: resolved

src/OpenTracing/Log/Fields.cs
ndrwrbgs: resolved

src/OpenTracing/Propagation/TextMap.cs
ndrwrbgs: resolved

src/OpenTracing/Propagation/TextMapExtractAdapter.cs
ndrwrbgs: resolved

src/OpenTracing/Propagation/TextMapInjectAdapter.cs
ndrwrbgs: resolved

src/OpenTracing/Tag/AbstractTag.cs
ndrwrbgs: resolved

src/OpenTracing/References.cs
ndrwrbgs: resolved

src/OpenTracing/Tag/Tags.cs
ndrwrbgs: resolved

src/OpenTracing/Tag/Tags.cs
ndrwrbgs: updated text to match the semantic_conventions spec. Planned to do this across the board after the port from java before

src/OpenTracing/Tag/Tags.cs
ndrwrbgs: resolved

src/OpenTracing/Tag/Tags.cs
ndrwrbgs: resolved

src/OpenTracing/UnsupportedFormatException.cs
ndrwrbgs: resolved

src/OpenTracing/ISpan.cs
ndrwrbgs: My implementation appears to use a reference. Method would be safer as far as an API w/o implementations goes as it supports both and leaves the caching decision up to the caller, but if you still feel property is the way to go I am willing to change.

src/OpenTracing/ISpan.cs
ndrwrbgs: Agreed, reverting all IDictionaries until #44 is resolved now that you're back around so that can get traction :)

ndrwrbgs:
@cwe1ss I reverted some things, like going back to the IDictionary for Log to match the java `Map` interface. Perhaps we should have IEnumerable<KeyValuePair>` but I'd rather have each deviation we make from the Java impl be reviewed in-and-of itself rather than lumped with this port.

If you don't want this to go into master, can you make a feature branch so we can start with this and build off of it with language-specific features we want (like perhaps extension methods etc).

Additionally, that would enable us to add the -utils folder in parallel/a separate review from this one, which would help us stay organized. TL;DR I'd suggest we consider this port 'done' but the actual C# implementation a WIP and iterate in a non-master branch until release in 1-2 weeks as you targetted.

src/OpenTracing/ISpan.cs
```diff
@@ -1,164 +1,133 @@
...
-        ///     { "event", "soft error" },
-        ///     { "type", "cache timeout" },
-        ///     { "waited.millis", 1500 }
-        /// });
-        /// </code>
-        /// </example>
-        ISpan Log(DateTimeOffset timestamp, IEnumerable<KeyValuePair<string, object>> fields);
+        /// <returns>The Span, for chaining</returns>
+        /// <seealso cref="Log(long, string)"/>
+        ISpan Log(long timestampMicroseconds, IDictionary<string, object> fields);
```
cwe1ss: Please keep the usage of DateTimeOffset everywhere. That difference was discussed in #4 and #42 and makes a lot of sense for .NET.

src/OpenTracing/LogFields.cs
```diff
@@ -1,45 +1,41 @@
...
-        /// <summary>
-        /// The type or "kind" of an error (only for event="error" logs).
-        /// <para />E.g., "Exception", "OSError".
-        /// </summary>
+        /// <summary>The type or "kind" of an error (only for event="error" logs). E.g., "Exception", "OSError"</summary>
         public const string ErrorKind = "error.kind";
 
         /// <summary>
-        /// The actual <see cref="Exception"/> object instance.
+        /// The actual Throwable/Exception/Error object instance itself. E.g., A java.lang.UnsupportedOperationException
```
cwe1ss: I think the existing comment (`/// The actual <see cref="Exception"/> object instance.`) makes more sense here since it is a better explanation for .NET users and doesn't mention a java type.

src/OpenTracing/Propagation/TextMap.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace OpenTracing.Propagation
+{
+    /// <summary>
+    /// <see cref="TextMap"/> is a built-in carrier for <see cref="ITracer.Inject{TCarrier}"/> and
+    /// <see cref="ITracer.Extract{TCarrier}"/>. TextMap implementations allows Tracers to read and write key:value String
+    /// pairs from arbitrary underlying sources of data.
+    /// </summary>
+    /// <seealso cref="ITracer.Inject{TCarrier}"/>
+    /// <seealso cref="ITracer.Extract{TCarrier}"/>
+    public interface TextMap : IEnumerable<KeyValuePair<string, string>>
```
cwe1ss: ITextMap to stay with .NET naming conventions

src/OpenTracing/LogFields.cs
ndrwrbgs: Will change but want to bring onto your radar my logic here:

I was concerned about that since tracing is distributed. I do not want to suggest that it will always actually be an Exception as if the distributed system has multiple languages it might not be.

However, this is log and not Baggage/Tag, and while it's possible an implementation put all it's Logs as Baggage (some... very test-oriented implementation :) ) in retrospect it's unlikely this would be a problem.

src/OpenTracing/ISpan.cs
ndrwrbgs: Will change but want to bring onto your radar my logic here:

So I was even involved in one of those I think, but in the past few months have been focused on perf at work. I have a minor concern regarding it here as DateTimeOffset is less performant than a long (e.g. to make the DateTimeOffset the caller has to parse from the clock into this struct anyway, and then if the tracing system does NOT want to use that first hand - which is quite possible given the java APIs expect long which is more cross-language - they'd have to parse it back to a long, a round trip that isn't strictly necessary.

That said, we can cross this bridge when we come to it (e.g. after we have some real implementations that have all other perf issues fixed, as it's unlikely this will be the bottleneck in any implementation at first)

cwe1ss: There isn't even a way to get the ticks without going through a DateTime(Offset), is there?

cwe1ss:
Thanks again for the changes! TextMap->ITextMap is still missing and there are some broken XML comments:

```
LogFields.cs(14,52): error CS1570: XML comment has badly formed XML -- '/' [/home/travis/build/opentracing/opentracing-csharp/src/OpenTracing/OpenTracing.csproj]
LogFields.cs(16,15): error CS1570: XML comment has badly formed XML -- 'summary' [/home/travis/build/opentracing/opentracing-csharp/src/OpenTracing/OpenTracing.csproj]
LogFields.cs(17,1): error CS1570: XML comment has badly formed XML -- 'summary' [/home/travis/build/opentracing/opentracing-csharp/src/OpenTracing/OpenTracing.csproj]
```

src/OpenTracing/ISpan.cs
```diff
@@ -1,164 +1,135 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 
 namespace OpenTracing
 {
+    using System;
```
cwe1ss: should be outside of the namespace as well :)

src/OpenTracing/Propagation/TextMapInjectAdapter.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        public IEnumerator<KeyValuePair<string, string>> GetEnumerator()
+        {
+            throw new InvalidOperationException(
+                $"{nameof(TextMapInjectAdapter)} should only be used with {nameof(ITracer)}.{nameof(ITracer.Inject)}");
+        }
+
+        public void Set(string key, string value)
+        {
+            _dictionary.Add(key, value);
```
cwe1ss: This should be `_dictionary[key] = value;` to ensure existing values are overwritten if it is called multiple times. Java's `put` has the same behavior.

src/OpenTracing/ISpan.cs
ndrwrbgs: acck I had to reset my settings for work, this should be easier :-)

src/OpenTracing/OpenTracing.csproj
```diff
@@ -6,6 +6,9 @@
 
 For the time being, mild backwards-incompatible changes may be made without changing the major version number. As OpenTracing and opentracing-csharp mature, backwards compatibility will become more of a priority.</Description>
     <PackageTags>opentracing;distributed-tracing;tracing;logging</PackageTags>
+    
+    <DocumentationFile>bin\$(Configuration)\netstandard1.0\OpenTracing.xml</DocumentationFile>
```
cwe1ss: This is already configured in Directory.Build.props but I only enabled it for Release-builds. I guess that's an unnecessary optimization so if you want, you can remove the `and '$(Configuration)'=='Release'` from there. I can also do it after this PR has been merged since it's not a blocker.

cwe1ss:
merged. @ndrwrbgs thanks again very much for this PR!

src/OpenTracing/OpenTracing.csproj
cwe1ss: I fixed this in the master-branch. xml files are now always created for all projects.

PR #51, opened 2017-11-14
Target both netstandard1.0 and net45 to remove unneeded dependencies

bmermet:
When it is installed on a project targeting the full .Net framework, the `Opentracing` NuGet package because of its dependency on `NETStandard.Library` pulls a lot of dependencies that are not needed. It has no impact on the resulting executable however it requires downloading a lot of packages for nothing and when used with the old csproj format it clutters the `packages.config` with all the .net core packages. For example installing Opentracing in an empty project produces the following `packages.config` file:

```xml
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="Microsoft.NETCore.Platforms" version="1.1.0" targetFramework="net461" />
  <package id="Microsoft.Win32.Primitives" version="4.3.0" targetFramework="net461" />
  <package id="NETStandard.Library" version="1.6.1" targetFramework="net461" />
  <package id="OpenTracing" version="0.10.0" targetFramework="net461" />
  <package id="System.AppContext" version="4.3.0" targetFramework="net461" />
  <package id="System.Collections" version="4.3.0" targetFramework="net461" />
  <package id="System.Collections.Concurrent" version="4.3.0" targetFramework="net461" />
  <package id="System.Console" version="4.3.0" targetFramework="net461" />
  <package id="System.Diagnostics.Debug" version="4.3.0" targetFramework="net461" />
  <package id="System.Diagnostics.DiagnosticSource" version="4.3.0" targetFramework="net461" />
  <package id="System.Diagnostics.Tools" version="4.3.0" targetFramework="net461" />
  <package id="System.Diagnostics.Tracing" version="4.3.0" targetFramework="net461" />
  <package id="System.Globalization" version="4.3.0" targetFramework="net461" />
  <package id="System.Globalization.Calendars" version="4.3.0" targetFramework="net461" />
  <package id="System.IO" version="4.3.0" targetFramework="net461" />
  <package id="System.IO.Compression" version="4.3.0" targetFramework="net461" />
  <package id="System.IO.Compression.ZipFile" version="4.3.0" targetFramework="net461" />
  <package id="System.IO.FileSystem" version="4.3.0" targetFramework="net461" />
  <package id="System.IO.FileSystem.Primitives" version="4.3.0" targetFramework="net461" />
  <package id="System.Linq" version="4.3.0" targetFramework="net461" />
  <package id="System.Linq.Expressions" version="4.3.0" targetFramework="net461" />
  <package id="System.Net.Http" version="4.3.0" targetFramework="net461" />
  <package id="System.Net.Primitives" version="4.3.0" targetFramework="net461" />
  <package id="System.Net.Sockets" version="4.3.0" targetFramework="net461" />
  <package id="System.ObjectModel" version="4.3.0" targetFramework="net461" />
  <package id="System.Reflection" version="4.3.0" targetFramework="net461" />
  <package id="System.Reflection.Extensions" version="4.3.0" targetFramework="net461" />
  <package id="System.Reflection.Primitives" version="4.3.0" targetFramework="net461" />
  <package id="System.Resources.ResourceManager" version="4.3.0" targetFramework="net461" />
  <package id="System.Runtime" version="4.3.0" targetFramework="net461" />
  <package id="System.Runtime.Extensions" version="4.3.0" targetFramework="net461" />
  <package id="System.Runtime.Handles" version="4.3.0" targetFramework="net461" />
  <package id="System.Runtime.InteropServices" version="4.3.0" targetFramework="net461" />
  <package id="System.Runtime.InteropServices.RuntimeInformation" version="4.3.0" targetFramework="net461" />
  <package id="System.Runtime.Numerics" version="4.3.0" targetFramework="net461" />
  <package id="System.Security.Cryptography.Algorithms" version="4.3.0" targetFramework="net461" />
  <package id="System.Security.Cryptography.Encoding" version="4.3.0" targetFramework="net461" />
  <package id="System.Security.Cryptography.Primitives" version="4.3.0" targetFramework="net461" />
  <package id="System.Security.Cryptography.X509Certificates" version="4.3.0" targetFramework="net461" />
  <package id="System.Text.Encoding" version="4.3.0" targetFramework="net461" />
  <package id="System.Text.Encoding.Extensions" version="4.3.0" targetFramework="net461" />
  <package id="System.Text.RegularExpressions" version="4.3.0" targetFramework="net461" />
  <package id="System.Threading" version="4.3.0" targetFramework="net461" />
  <package id="System.Threading.Tasks" version="4.3.0" targetFramework="net461" />
  <package id="System.Threading.Timer" version="4.3.0" targetFramework="net461" />
  <package id="System.Xml.ReaderWriter" version="4.3.0" targetFramework="net461" />
  <package id="System.Xml.XDocument" version="4.3.0" targetFramework="net461" />
</packages>
```
To avoid this I suggest to target both netstandard1.0 and net45, that way when used with the full framework the net45 dll will be used and will not pull all these dependencies. The only downside I see with this approach is that AFAIK it prevents building the package on non Windows platforms. This will however have no consequences on non Windows users of the package, since the package will still be compatible with netstandard1.0.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -26,10 +26,10 @@ echo
 echo dotnet-build
 echo ----------------------
 
-dotnet build -c $CONFIGURATION --no-incremental
+dotnet build -c $CONFIGURATION --no-incremental /p:NonWindowsBuild=true
 
 echo
 echo dotnet-test
 echo ----------------------
 
-dotnet test test/OpenTracing.Tests/OpenTracing.Tests.csproj -c $CONFIGURATION --no-build
+dotnet test test/OpenTracing.Tests/OpenTracing.Tests.csproj -c $CONFIGURATION --no-build /p:NonWindowsBuild=true
--- a/src/OpenTracing/OpenTracing.csproj
+++ b/src/OpenTracing/OpenTracing.csproj
@@ -8,4 +8,8 @@ For the time being, mild backwards-incompatible changes may be made without chan
     <PackageTags>opentracing;distributed-tracing;tracing;logging</PackageTags>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(NonWindowsBuild)' == 'true' ">
+      <TargetFrameworks>netstandard1.0;</TargetFrameworks>
+  </PropertyGroup>
+
 </Project>
--- a/src/OpenTracing/OpenTracing.csproj
+++ b/src/OpenTracing/OpenTracing.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.0</TargetFramework>
+    <TargetFrameworks>netstandard1.0;net45</TargetFrameworks>
     <Description>This library is a .NET implementation of the OpenTracing API. To fully understand this platform API, it's helpful to be familiar with the OpenTracing project and terminology more generally.
 
 For the time being, mild backwards-incompatible changes may be made without changing the major version number. As OpenTracing and opentracing-csharp mature, backwards compatibility will become more of a priority.</Description>
--- a/test/OpenTracing.Tests/OpenTracing.Tests.csproj
+++ b/test/OpenTracing.Tests/OpenTracing.Tests.csproj
@@ -4,6 +4,10 @@
     <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(NonWindowsBuild)' == 'true' ">
+      <TargetFrameworks>netcoreapp1.1;</TargetFrameworks>
+  </PropertyGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\..\src\OpenTracing\OpenTracing.csproj" />
   </ItemGroup>
--- a/test/OpenTracing.Tests/OpenTracing.Tests.csproj
+++ b/test/OpenTracing.Tests/OpenTracing.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.1</TargetFramework>
+    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
   </PropertyGroup>
 
   <ItemGroup>


bmermet:
Hi guys,
My change broke the CI since it's running on Linux, so I have added additional conditions in the build to make it build and run on Linux with a flag.

src/OpenTracing/OpenTracing.csproj
```diff
@@ -1,11 +1,15 @@
...
   <PropertyGroup>
-    <TargetFramework>netstandard1.0</TargetFramework>
+    <TargetFrameworks>netstandard1.0;net45</TargetFrameworks>
     <Description>This library is a .NET implementation of the OpenTracing API. To fully understand this platform API, it's helpful to be familiar with the OpenTracing project and terminology more generally.
 
 For the time being, mild backwards-incompatible changes may be made without changing the major version number. As OpenTracing and opentracing-csharp mature, backwards compatibility will become more of a priority.</Description>
     <PackageTags>opentracing;distributed-tracing;tracing;logging</PackageTags>
   </PropertyGroup>
 
+  <PropertyGroup Condition=" '$(NonWindowsBuild)' == 'true' ">
```
TerribleDev: can't this just be `'$(OS)' == 'Windows'` since msbuild injects the variable and then you don't have to pass in the proper cli args

build.sh
```diff
@@ -26,10 +26,10 @@ echo
 echo dotnet-build
 echo ----------------------
 
-dotnet build -c $CONFIGURATION --no-incremental
+dotnet build -c $CONFIGURATION --no-incremental /p:NonWindowsBuild=true
```
TerribleDev: you can use the msbuild that ships with mono to compile binaries for .net 45

src/OpenTracing/OpenTracing.csproj
bmermet: Thanks ! I didn't know this variable existed, it is simpler that way indeed.

cwe1ss:
Which version of the .NET framework are you using? Ideally, I'd like to change the TargetFramework to netstandard20 anyway since that's what MS recommends (at least if I understand correctly) and it also no longer references that package. It only supports .NET 4.6.1+ though.

If there's people who are still targeting .NET 4.5 we can of course target netstandard2.0;net45 (or net452 which is the only supported version)

ndrwrbgs:
Work is very very slow to move so we're still on 4.5. I had to move parts of the code to 4.6 to support my local build of this, but it'd be nice if we could support 4.5 for ~1 more year or so, if possible!

cwe1ss:
I implemented this with the mentioned msbuild changes in #56 since there were no further commits here. hope that's ok for you @bmermet ?! I can also undo that in my PR if you have time to fix this PR in the next few days.

bmermet:
All good for me, I'll close this PR then.

PR #59, opened 2018-02-09
finishSpanOnDispose instead of finishSpanOnClose

cwe1ss:
Java has `Closable ` with a `close` method and .NET has `IDisposable` with a `Dispose` method.

To reduce confusion, I think it therefore makes sense to rename the parameter `finishSpanOnClose` to `finishSpanOnDispose`.


cwe1ss:

carlosalberto:
@cwe1ss It looks good to me.

PR #60, opened 2018-02-09
EditorConfig coding conventions (copied from dotnet/corefx)

cwe1ss:
The `.editorconfig` file contains coding conventions and should prevent contributors from sending PRs with different styles. 

It was copied from the [dotnet/corefx](https://github.com/dotnet/corefx) repository so it's "as .NET as it gets". I know that we could debate about this endlessly but I hope that copying the conventions from one of the biggest .NET OSS projects is OK for everyone.

The .editorconfig file is supported out of the box by Visual Studio 2017 and there's [a plugin](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) for Visual Studio Code.

It suggests that each text file ends with \n which [apparently is a good thing](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file) so I've fixed that in every file as well.


cwe1ss:
There's also [a plugin for VS 2015](https://marketplace.visualstudio.com/items?itemName=EditorConfigTeam.EditorConfig) and [R# 2017.1+ supports it as well](https://www.jetbrains.com/help/resharper/Using_EditorConfig.html). Not sure though how many of the settings they support. Most of it is suggestions anyway. The most important things which should be enforced everywhere are indentations, sorted usings etc.

cwe1ss:

PR #61, opened 2018-02-10
GlobalTracer ported from Java

### LINKED ISSUE #45: Add a static Trace Class
Add a static Trace class with the following properties:

```
public static class Trace
{
   static ITracer tracer = new NullTracer();

   public static ITracer Tracer
   {
      get { return tracer; }
      set { tracer = value ?? new NullTracer(); }
}
```

The ```Tracer``` property is initialized during application startup and provides entry point for accessing the tracer. This pattern is similar with those provided by many loggers e.g. **Serilog** has something akin to:

```
public class Log
{
  public ILogger Logger { get; set; }
}

```

cwe1ss:
This is a port from the `opentracing-java` project. It currently has the following deviations:

* C# does not have the `synchronized` keyword and usage of `MethodImpl.Synchronized` [is not really recommended](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.methodimplattributes?view=netframework-4.7.1#System_Reflection_MethodImplAttributes_Synchronized) so I'm using regular locking instead. I might have messed that up so any feedback is welcome. :)
* Java's `tracer` variable is static. I do not really know why as `GlobalTracer` is more or less a regular wrapper/decorator type so it should be an instance variable. 

Note that this currently does not compile as it requires the NoopTracer from #57. That's also the reason why there's no unit tests yet.

This will resolve #45.


cwe1ss:

src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        /// <summary>
+        /// Singleton instance.
+        /// <para/>
+        /// Since we cannot prevent people using <see cref="Instance"/> as a constant,
+        /// this guarantees that references obtained before, during or after initialization
+        /// all behave as if obtained <em>after</em> initialization once properly initialized.<br/>
+        /// As a minor additional benefit it makes it harder to circumvent the <see cref="ITracer"/> API.
+        /// </summary>
+        private static readonly GlobalTracer s_instance = new GlobalTracer();
```
ndrwrbgs: Should GlobalTracer have a private ctor since it appears it's meant to be a singleton?

src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// Identify whether a <see cref="ITracer"/> has previously been registered.
+        /// <para/>
+        /// This check is useful in scenarios where more than one component may be responsible
+        /// for registering a tracer.
+        /// </summary>
+        /// <returns>Whether a tracer has been registered.</returns>
+        public static bool IsRegistered()
+        {
+            return !(s_instance._tracer is NoopTracer);
```
ndrwrbgs: [nit] Slightly misleading as the user could have registered the `NoopTracer`. I'd rather set a variable than rely on the type.

src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -0,0 +1,125 @@
...
+                    throw new InvalidOperationException("There is already a current global Tracer registered.");
+
+                s_instance._tracer = tracer;
+            }
+        }
+
+        private ITracer _tracer = NoopTracerFactory.Create();
+
+        // <inheritdoc/>
+        public IScopeManager ScopeManager => _tracer.ScopeManager;
```
ndrwrbgs: Given this isn't used in this class seems to suggest we are doing more than SRP, which I'd argue we are. We're trying to wrap a tracer while also trying to provide singleton global access.  Might I suggest something more like

```C#
public static class GlobalTracer/TracerProvider
{
  public static ITracer Tracer { get; set; } /* more details about how to implement this addressing the relevant concerns in another comment */
}
```

src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// Returns the constant <see cref="GlobalTracer"/>.
+        /// <para/>
+        /// All methods are forwarded to the currently configured tracer.<br/>
+        /// Until a tracer is explicitly configured via <see cref="Register(ITracer)"/>,
+        /// the <see cref="NoopTracer"/> is used.
+        /// </summary>
+        /// <returns>The global tracer constant.</returns>
+        /// <seealso cref="Register(ITracer)"/>
+        public static ITracer Instance => s_instance;
```
ndrwrbgs: Minor concern, but could add up.
Since this is itself exposed as an `ITracer` rather than it's real type, it will result in all calls to it like `BuildSpan` becoming `CallVirt`'s. Then, since it wraps another, it has another `CallVirt`.

I really like information hiding, but in this case exposing it's real type may be best since it's likely all tracing will come through this class.

src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// Register a <see cref="ITracer"/> to back the behaviour of the global tracer (<see cref="Instance"/>).
+        /// <para/>
+        /// Registration is a one-time operation, attempting to call it more often will result in a runtime exception.
+        /// <para/>
+        /// Every application intending to use the global tracer is responsible for registering it once
+        /// during its initialization.
+        /// </summary>
+        /// <param name="tracer">Tracer to use as global tracer.</param>
+        public static void Register(ITracer tracer)
```
ndrwrbgs: It's possible a developer at the composition layer wants to register one tracer but someone wants to wrap that for composition (or they have a InitializeStandardTracer in a base class and then some custom wrapping they want). The current code prevents that.
A possible way to handle it is

```C#
public delegate ITracer ExchangeTracerDelegate(ITracer existingTracer);

public static ITracer ExchangeTracer([NotNull] ExchangeTracerDelegate newTracerDelegate) // Oops, my love for nullability annotations from ReSharper is showing :)
{
    lock (s_lock)
    {
        s_instance._tracer = newTracerDelegate(s_instance._tracer);
    }
}
```

ndrwrbgs: Note we also do not support someone saying "I want to intercept trace messages from my callee here and down", which would require AsyncLocal for the tracer.
While I'd like to support that, I wouldn't do so until I had measured the performance of AsyncLocal.

src/OpenTracing/Util/GlobalTracer.cs
cwe1ss: yes, thank you! will fix it!

src/OpenTracing/Util/GlobalTracer.cs
cwe1ss: That's a feature. Having a "NoopTracer" by definition means that no tracer is set.

src/OpenTracing/Util/GlobalTracer.cs
cwe1ss: The problem with this is that people could save that `TracerProvider.Tracer` variable locally and therefore never get a new tracer if someone replaces it. That's the reason why in this class, `GlobalTracer.Instance` always returns the one singleton `GlobalTracer`.

The instance types are not used because GlobalTracer is just a simple decorator/wrapper type.

src/OpenTracing/Util/GlobalTracer.cs
cwe1ss: Hmmm... no idea what the perf impact is here. I'll do a little bit of research. Returning `GlobalTracer` probably means that IntelliSense no longer shows documentation since `<inheritdoc/>` is not really supported in IDEs AFAIK. We would have to copy all the documentation.

src/OpenTracing/Util/GlobalTracer.cs
cwe1ss: These are requirements that are not yet formally specified (https://github.com/opentracing/specification/issues/9 [multiple tracers], https://github.com/opentracing/specification/issues/76 [observers]) so they are no-goals of this implementation.

src/OpenTracing/Util/GlobalTracer.cs
carlosalberto: > The instance types are not used because GlobalTracer is just a simple decorator/wrapper type.

Correct, this is the expected behavior :)

src/OpenTracing/Util/GlobalTracer.cs
carlosalberto: @ndrwrbgs That's an interesting feature. I'd personally recommend you post an issue against the Java repo and have people discuss it ;)

src/OpenTracing/Util/GlobalTracer.cs
carlosalberto: It looks like something that could be done afterwards as an optimization (an issue around this could be created so it doesn't get forgotten, and eventually discussed/handled).

Issue #62, opened 2018-02-10
Target netstandard1.3 and net45


cwe1ss:
I'd like to make sure people don't do weird stuff with this library unless we know about it. One way to ensure this is to only target officially supported versions of .NET Core and the full .NET framework.

We currently target netstandard1.0 but this also includes stuff like Silverlight and Windows Phone, which are no longer supported by Microsoft.

For this reason, I'd like to target `netstandard1.3` (which supports .NET Core 1+, .NET 4.6+, Xamarin, Mono, UWP) and `net452` to support .NET 4.5.2 (which is the only officially supported .NET 4.5 version)

See [.NET Standard](https://docs.microsoft.com/en-us/dotnet/standard/net-standard) for details.

If people really need it for older platforms, we can always talk about it and loosen this requirement - it's always easier to add older targets than to remove them.

Edit 2018-02-20: We are using `net45` instead of `net452` (see comments)

cwe1ss:
@bmermet @ndrwrbgs you've said in #51 that you're targeting .NET 4.5 at work. Are you targeting .NET 4.5.2 or an older version? (Note that older versions are no longer supported by Microsoft)

cwe1ss:
I've committed this change with #63. Feel free to comment here if that's a problem... we can easily change that.

MikeGoldsmith:
I think we should target net45 too. It's still listed as part of the compatibility matrix on the MS documentation pages and because we fall back to CallContext when using < net462 (AsyncLocal not available) we're still covered.

I would also require it as part of my integration with Couchbase SDK as we currently target net45, netstandard1.3 and netstandard2.0. As an integration library, we're not in control of what framework version our customers use.

cwe1ss:
I changed it to `net45` instead of `net452`.

cwe1ss:
I've just released v0.11.0. Please create a new issue with your scenario if you need support for lower .NET Standard versions.

Issue #55, opened 2018-02-01
0.11 Release plan


cwe1ss:
This is a meta issue to track/discuss what should be done in the next release. The tasks reflect my proposal, any feedback/help is more than welcome!

### Parity with opentracing-api from Java

The main thing that's missing right now is the support for scopes. This will only add new types/members but it will still be a breaking change (for implementations) since it adds members to an existing interface (ITracer).

Apart from that, the C# version currently has some other differences. We want to be as close to the Java version as possible so we will only keep differences that make sense for C#/.NET. We will do this with #50 which basically replaces the whole codebase and is based on the current master of opentracing-java.

This PR will introduce the following changes to the current code:
* Log methods will use `IDictionary` instead of `IEnumerable`. There's a discussion in #44 about whether or not there should be other changes regarding this in a later version.
* We will remove the `FollowsFrom()` methods in `ISpanBuilder` as they currently don't exist in Java.
* We will use Java's Tag types instead of our simple Tags-type with constants.
* We will use Java's propagation types. Our types were slightly different.
* ~~We will add support for the BINARY format. It will be based on `Stream`~~
  * ~~I still don't know if that's correct since I don't understand how this format is used - see also #7.~~
  * Edit: we decided not to add BINARY for now as there's also some discussion around it in Java.
* We will remove the `NullTracer` from the core API. We will follow Java's lead and add it as a `NoopTracer` to a different package.
  * The current implementation is not good as there's currently a namespace and a type called `NullTracer` so one has to write `NullTracer.NullTracer`. 
* We will remove our exception types (`OpenTracingException`, `UnsupportedFormatException`) as they don't exist in Java (they were from an unmerged PR)

Changing these things obviously results in a ton of breaking changes but I guess the impact is still minimal and the benefit of being on par with Java outweighs it.

We will keep the following differences. __Please comment if you want to remove any of these differences as well!__

* We are using `DateTimeOffset` instead ticks for tracking timestamps. The reasons for this are described in #4 and #42.
* We have a `OpenTracing.LogFields` type and Java has a `OpenTracing.Log.Fields` type, so the namespace and type names are different. 
  * I think that makes a lot more sense as it doesn't introduce a new namespace for one file. Also, the Log-method is in the "OpenTracing" namespace so there's no point in having the constants in a different namespace.

*Tasks:*
- [x] Merge #50
- [x] Remove exception types #64 

### Enhancements / Additions

[...]

PR #63, opened 2018-02-10
AsyncLocal based scope manager (ported from Java)

cwe1ss:
This is a port from opentracing-java's ThreadLocalScopeManager. 

* For storage it either uses `AsyncLocal` (.NET 4.6+, .NET Core) or the `CallContext` (.NET 4.5). This code is based on ASP.NET Core's [HttpContextAccessor](https://github.com/aspnet/HttpAbstractions/blob/622d112372ed11075af160b1fb5a11a7637ff516/src/Microsoft.AspNetCore.Http/HttpContextAccessor.cs).
* Since I had to fiddle around with #if-defs I'm now targeting `netstandard1.3` and `net452` as described in #62 - I'll change the targets here should that issue decide differently.
* The unit tests are also ported from Java.
* I also had to slightly change the build script (as the tests now target multiple frameworks) and I've updated the nuget dependencies.
* I have NOT ported Java's `AutoFinishScopeManager` as I'm not sure if that's required in .NET (due to async/await). I'd say we port that later if someone really needs it.
* The project structure assumes only one OpenTracing package - I'll change this as well should we decide differently in #55.

### DIFF
--- a/src/OpenTracing/Util/AsyncLocalScope.cs
+++ b/src/OpenTracing/Util/AsyncLocalScope.cs
@@ -1,0 +1,43 @@
+namespace OpenTracing.Util
+{
+    /// <summary>
+    /// The <see cref="AsyncLocalScope"/> is a simple <see cref="IScope"/> implementation
+    /// that relies on C#'s AsyncLocal/CallContext storage primitive.
+    /// </summary>
+    /// <seealso cref="IScopeManager"/>
+    public class AsyncLocalScope : IScope
+    {
+        private readonly AsyncLocalScopeManager _scopeManager;
+        private readonly ISpan _wrappedSpan;
+        private readonly bool _finishOnDispose;
+        private readonly IScope _scopeToRestore;
+
+        public AsyncLocalScope(AsyncLocalScopeManager scopeManager, ISpan wrappedSpan, bool finishOnDispose)
+        {
+            _scopeManager = scopeManager;
+            _wrappedSpan = wrappedSpan;
+            _finishOnDispose = finishOnDispose;
+
+            _scopeToRestore = scopeManager.Active;
+            scopeManager.Active = this;
+        }
+
+        public ISpan Span => _wrappedSpan;
+
+        public void Dispose()
+        {
+            if (_scopeManager.Active != this)
+            {
+                // This shouldn't happen if users call methods in the expected order. Bail out.
+                return;
+            }
+
+            if (_finishOnDispose)
+            {
+                _wrappedSpan.Finish();
+            }
+
+            _scopeManager.Active = _scopeToRestore;
+        }
+    }
+}
--- a/test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
+++ b/test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
@@ -1,0 +1,80 @@
+﻿using NSubstitute;
+using OpenTracing.Util;
+using Xunit;
+
+namespace OpenTracing.Tests.Util
+{
+    public class AsyncLocalScopeManagerTests
+    {
+        private AsyncLocalScopeManager _source;
+
+
+        public AsyncLocalScopeManagerTests()
+        {
+            _source = new AsyncLocalScopeManager();
+        }
+
+        [Fact]
+        public void MissingActiveScope()
+        {
+            IScope missingScope = _source.Active;
+            Assert.Null(missingScope);
+        }
+
+        [Fact]
+        public void DefaultActivate()
+        {
+            ISpan span = Substitute.For<ISpan>();
+
+            using (IScope scope = _source.Activate(span, false))
+            {
+                Assert.NotNull(scope);
+                IScope otherScope = _source.Active;
+                Assert.Same(otherScope, scope);
+            }
+
+            // Make sure the span is not finished.
+            span.DidNotReceive().Finish();
+
+            // And now it's gone:
+            IScope missingScope = _source.Active;
+            Assert.Null(missingScope);
+        }
+
+        [Fact]
+        public void FinishSpanOnDispose()
+        {
+            ISpan span = Substitute.For<ISpan>();
+
+            using (IScope scope = _source.Activate(span, true))
+            {
+                Assert.NotNull(scope);
+                Assert.NotNull(_source.Active);
+            }
+
+            // Make sure the ISpan got finish()ed.
+            span.Received(1).Finish();
+
+            // Verify it's gone.
+            Assert.Null(_source.Active);
+        }
+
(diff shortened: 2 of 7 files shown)


cwe1ss:

src/OpenTracing/Util/AsyncLocalScope.cs
```diff
@@ -0,0 +1,43 @@
...
+    /// </summary>
+    /// <seealso cref="IScopeManager"/>
+    public class AsyncLocalScope : IScope
+    {
+        private readonly AsyncLocalScopeManager _scopeManager;
+        private readonly ISpan _wrappedSpan;
+        private readonly bool _finishOnDispose;
+        private readonly IScope _scopeToRestore;
+
+        public AsyncLocalScope(AsyncLocalScopeManager scopeManager, ISpan wrappedSpan, bool finishOnDispose)
```
ndrwrbgs: [nit] it looks like the only place `wrappedSpan` is supplied here it's not wrapped in anything?

test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            // Verify it's gone.
+            Assert.Null(_source.Active);
+        }
+
+        [Fact]
+        public void DontFinishSpanNoDispose()
+        {
+            ISpan span = Substitute.For<ISpan>();
+
+            using (IScope scope = _source.Activate(span, false))
```
ndrwrbgs: [nit] I generally prefer named args for booleans (or nulls) because it makes the call site make a lot more sense.
E.g. in this place you'd be able to see we expect no `Finish()` because we said `finishOnDispose: false`
Without it you require more contextual brain-knowledge to be able to understand the test

src/OpenTracing/Util/AsyncLocalScope.cs
ndrwrbgs: Oooh it's the span that is BEING wrapped, not a span that IS wrapped. Obviously your naming is fine now that I realize this, but given I was confused for a bit suggests possibility for better naming (subjective, feel free to leave if you disagree)

src/OpenTracing/Util/AsyncLocalScope.cs
```diff
@@ -0,0 +1,43 @@
+namespace OpenTracing.Util
+{
+    /// <summary>
+    /// The <see cref="AsyncLocalScope"/> is a simple <see cref="IScope"/> implementation
+    /// that relies on C#'s AsyncLocal/CallContext storage primitive.
+    /// </summary>
+    /// <seealso cref="IScopeManager"/>
+    public class AsyncLocalScope : IScope
```
ndrwrbgs: This relates to the other review comment I made regarding the APIs. Using this structure removes the ability for any implementation to expect that the Scope they get back will be the one they pushed out of their APIs, which can cause some weird issues.

Do not have a good suggestion at this time, so this is not a comment to request change but just to start your brain pondering over solutions.

src/OpenTracing/Util/AsyncLocalScope.cs
```diff
@@ -0,0 +1,43 @@
...
+            scopeManager.Active = this;
+        }
+
+        public ISpan Span => _wrappedSpan;
+
+        public void Dispose()
+        {
+            if (_scopeManager.Active != this)
+            {
+                // This shouldn't happen if users call methods in the expected order. Bail out.
```
ndrwrbgs: Throw `InvalidOperationException`?

```
The exception that is thrown when a method call is invalid for the object's current state
```

ndrwrbgs: Depends on your views... Throw from Dispose is generally frowned upon, but which is worse:
1) Failing silently leaving a bug to go potentially unnoticed for quite some time and hard to debug
2) Throwing from Dispose on code errors

The reason Throw on Dispose is frowned upon is because the following
```C#
try
{
 using (var span = /* ... */)
 {
   // Do stuff
   throw new IOException();
 }
}
catch (IOException)
{
 // Not hit if (non-IOException) thrown from Dispose, and the exception being thrown is not available to `Dispose` so it can't even be wrapped
}
```

But I generally feel that's okay because you only loose an exception if you're 1) already in an exceptional state and 2) have done something exceptional with the library that suggests a developer bug (I avoid it when it's a state-based/concurrency-based bug but when it's something that's likely just a coding error it seems better to fail-fast to me)

src/OpenTracing/Util/AsyncLocalScope.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly bool _finishOnDispose;
+        private readonly IScope _scopeToRestore;
+
+        public AsyncLocalScope(AsyncLocalScopeManager scopeManager, ISpan wrappedSpan, bool finishOnDispose)
+        {
+            _scopeManager = scopeManager;
+            _wrappedSpan = wrappedSpan;
+            _finishOnDispose = finishOnDispose;
+
+            _scopeToRestore = scopeManager.Active;
```
ndrwrbgs: Do we need to explicitly set the `Parent` here? I notice my InMemory implementation also does not (at this location) but I'm curious if we both missed part of the spec

src/OpenTracing/Util/AsyncLocalScope.cs
cwe1ss: Naming is from Java, except that I named it `wrappedSpan` instead of just `wrapped`.

test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
cwe1ss: fixed

src/OpenTracing/Util/AsyncLocalScope.cs
cwe1ss: AFAIK not throwing an exception here has been a deliberate decision in Java as tracers are not meant to break the application (in the same way that loggers are not meant to break it). So if you want to have this changed we should discuss it in a separate issue.

src/OpenTracing/Util/AsyncLocalScope.cs
cwe1ss: Could you clarify this? I do not understand what you mean.

src/OpenTracing/Util/AsyncLocalScope.cs
cwe1ss: Could you clarify this? I do not understand what you mean.

ndrwrbgs:
Personally, I think wrapped may be a little less misleading :) but
non-issue, it didn’t waste too much time reparsing.

PR #64, opened 2018-02-10
Remove exception types `OpenTracingException` and `UnsupportedFormatException`

cwe1ss:
We currently have the types `OpenTracingException` and `UnsupportedFormatException`. However, they do not exist in Java.

There has been a discussion about this in https://github.com/opentracing/opentracing-java/issues/42 and an open PR https://github.com/opentracing/opentracing-java/pull/43 by the time I've added these types. However, the PR has never been merged and I've never removed the types.

Removing them would be another step in being closer to the Java API.


ndrwrbgs:
Java has rich exception types that can be reused and checked. C# on the other hand suggests a custom exception type defined for each scenario. This change would prevent the library users from knowing what types of exceptions to handle, and prevent authors from knowing what types to throw.

ndrwrbgs:
To be less generic and more evidence based in my rationale I've performed the following:
Take the top 4 packages nuget.org (4 because 5 and 6 are non-C# code :) ), check how they handle exceptions.

| Package | Exception type | Link |
| ------------- |:-------------| -----|
| Newtonsoft.Json/Json.Net      | JsonException | [link](https://github.com/JamesNK/Newtonsoft.Json) |
| NUnit      | AssertionException  |   [link](https://github.com/nunit) |
| EntityFramework | EntityException |   [link](https://github.com/aspnet/EntityFramework6) |
| MySql | MySqlException | [link](https://dev.mysql.com/doc/dev/connector-net/6.10/html/T_MySql_Data_MySqlClient_MySqlException.htm) |

cwe1ss:
I understand your point. The thing is that in most cases, tracers are not supposed to break the application (in a same way that loggers are not supposed to). And the bigger isue is that, unfortunately, the specification currently does not really define the error behavior. There's just one open issue (https://github.com/opentracing/specification/issues/33) for how to handle missing/corrupt span contexts on extract. And there's the previously mentioned java issues that have never been resolved.

So the motivation here is to ensure the C# API only contains stuff that is formally defined as everything else is a deviation and problematic should the specification come to a different conclusion later. 

By releasing a version now that has feature-parity with Java, we can ensure that all of these problems can be discussed in a formal way and will be integrated into Java and C# in the same way.

I think that the in-process propagation took up most of the specification team's time last year so I think that now that this has been resolved and now that there's also this new cross-language meeting, we should be able to resolve these issues in an acceptable time frame.

Thoughts?

ndrwrbgs:
I agree, mostly. I would be cautious about releasing any version without
them though as it’s much easier to remove this constraint than add it later
(when some implementations will surely have their own exception types and
then we will have a mess).

I agree that it shouldn’t throw/break the application but specifying an
exception says IF you’re going to, at least do it this way; not that they
should/have to.

carlosalberto:
> By releasing a version now that has feature-parity with Java, we can ensure that all of these problems can be discussed in a formal way and will be integrated into Java and C# in the same way.

I agree on this. `opentracing-python`, for example, has their own `Exception`s precisely matching these ones - but it's a different environment altogether, and not quite modeled after the Java API, so I'd stick with what Java has **now** in this regard.

carlosalberto:
>  I would be cautious about releasing any version without
them though as it’s much easier to remove this constraint than add it later

So I think we are in a very early stage of adoption, which means we can still play around and experiment with stuff - even the Java API had this changes in its latest version.

(Of course, once we have a good, **stable** release, then we should stay uniform on what we expose ;) )

MikeGoldsmith:
I agree with @cwe1ss - I think we should remove these exception types because we do not have clear understanding when and how they should be used. Providing them at this point without clear intent can lead to misuse.

I do agree that when when we have defined error behaviour, custom exception types are the preferred approach in .NET and these exceptions are a good starting point.

cwe1ss:
I'll merge this since it seems like a majority is in favor of removing them for now.

PR #65, opened 2018-02-12
Move to Apache License 2.0

cwe1ss:
All CNCF projects should use the Apache License 2.0 - see https://github.com/opentracing/specification/issues/61#issuecomment-364682071 for a discussion.

As we haven't had a license header at the top of every file so far, I've not added them here either.

Please file any objections until 19th of February. 

/cc previous contributors: @dawallin @ndrwrbgs @tedsuo @bhs

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,201 @@
-The MIT License (MIT)
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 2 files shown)


bhs:
LGTM, thanks!

dawallin:
It's ok with me.

PR #66, opened 2018-02-12
Makes Tag setter public

cwe1ss:
#50 introduced Java's "tag" classes. In Java, the `AbstractTag` has a `protected set()` method and each derived class changes the visibility by having a `public set` method. #50 used `protected` everywhere which obviously doesn't allow one to actually use the classes.

It's not possible in C# to change the visibility of an overriden method so I made `AbstractTag`'s method `public`. I don't see any reason why it shouldn't be public and I don't understand why Java even made this decision.

I'll not mark this as a deviation as it's just a technical difference and not a feature difference.

As this clearly is a bug, I'll merge it tomorrow as well.

### DIFF
--- a/src/OpenTracing/Tag/AbstractTag.cs
+++ b/src/OpenTracing/Tag/AbstractTag.cs
@@ -9,6 +9,6 @@
 
         public string Key { get; }
 
-        protected abstract void Set(ISpan span, TTagValue tagValue);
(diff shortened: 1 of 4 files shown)


carlosalberto:
Makes sense to me (in addition to the inability to change the accessibility modifier).

Manigill:
07cacf87dc573412ecd69c3ee6f49b1bf63f71f8

PR #68, opened 2018-02-12
IFormat marker interface and public Format struct (alternative for #67)

cwe1ss:
Relates to #67 (read that first).

This would be my ideal solution. It's a mix of what we've had before (the public `Format` struct) and the `IFormat` marker interface.

Advantages:
* Tracers can use the `IFormat` marker interface to store lists of formats 
* Tracers can use the `Name` property to identify them by name
* The `Format` struct is public which means people who create their own formats don't have to duplicate the existing `BuiltinFormats.Builtin` struct. 
* We can reactive the format unit tests (which are currently pretty much useless due to `BuiltinFormats.Builtin` not being publicly available)

Maybe we should do this as we __have to__ deviate here anyway?!? 

Thoughts?

### DIFF
--- a/src/OpenTracing/Propagation/IFormat.cs
+++ b/src/OpenTracing/Propagation/IFormat.cs
@@ -1,5 +1,20 @@
-﻿namespace OpenTracing.Propagation
+﻿using System;
+
+namespace OpenTracing.Propagation
 {
+    /// <summary>
+    /// This is a marker interface to allow <see cref="Format{TCarrier}"/> instances being used without their type constraint.
+    /// </summary>
+    /// <seealso cref="ITracer.Inject{TCarrier}"/>
+    /// <seealso cref="ITracer.Extract{TCarrier}"/>
+    public interface IFormat
+    {
+        /// <summary>
+        /// The unique name for this format.
+        /// </summary>
+        string Name { get; }
+    }
+
     /// <summary>
     /// Format instances control the behavior of <see cref="ITracer.Inject{TCarrier}"/> and
     /// <see cref="ITracer.Extract{TCarrier}"/> (and also constrain the type of the carrier parameter to same). Most
@@ -12,7 +27,20 @@
     /// </summary>
     /// <seealso cref="ITracer.Inject{TCarrier}"/>
     /// <seealso cref="ITracer.Extract{TCarrier}"/>
-    public interface IFormat<TCarrier>
+    public struct Format<TCarrier> : IFormat
     {
+        /// <inheritdoc/>
+        public string Name { get; }
+
+        public Format(string name)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+        }
+
+        /// <summary>Short name for the formats as they tend to show up in exception messages</summary>
+        public override string ToString()
+        {
+            return $"Format.{Name}";
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/OpenTracing/Propagation/IFormat.cs
```diff
@@ -1,5 +1,20 @@
-﻿namespace OpenTracing.Propagation
+﻿using System;
+
+namespace OpenTracing.Propagation
 {
+    /// <summary>
+    /// This is a marker interface to allow <see cref="Format{TCarrier}"/> instances being used without their type constraint.
```
ndrwrbgs: What is the benefit of this? I see the storage of tracers in the Conversation pane, but how many formats do you expect a tracer would implement? And even if they implemented many, they'd have to go through all of those and convert them back to the real type anyway. Also, how a tracer stores the formats internally is their concern, could they not implement this themselves as they desire (e.g. if they had a bunch of custom formats that would actually make this worth it [I wouldn't bother with a collection until I had >4 types] couldn't they make an interface for their own types [perhaps not possible if libraries are exposing the types]).

But mostly, what does this give us over the user just storing them as `object` in their internal definition?

P.S. most of my recoil to this is motivated by 1) interface>struct is a pattern that often results in boxing, which makes the setup have worse performance than if it had just been interface>class, obviously this depends on usage though; 2) marker interfaces are believed by some (myself included) to be a code smell that. While it doesn't mean it's wrong, it triggers closer inspection of the use case. You can read from [NDepends](https://blog.ndepend.com/marker-interface-isnt-pattern-good-idea/) for their take on them, and whether you agree or not you'd hopefully come out understanding the 'other side' a bit more :) My biggest grievance with them is an interface with no methods isn't really an "interface".

cwe1ss: It's all about this line: https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/com/uber/jaeger/Tracer.java#L538

You must be able to register formats in a dynamic way with your tracer, so the tracer either has to store a list of format names or a list of format objects, both of which is not possible with the current design.

The list could be of type `object` of course, but that's rather weird/hacky as well IMO. 

What would be your preferred solution?

ndrwrbgs: I see the concern. I'd note here though

```java
    @SuppressWarnings("unchecked")
    <T> Injector<T> getInjector(Format<T> format) {
      return (Injector<T>) injectors.get(format);
    }

    @SuppressWarnings("unchecked")
    <T> Extractor<T> getExtractor(Format<T> format) {
      return (Extractor<T>) extractors.get(format);
    }

    public <T> void register(Format<T> format, Injector<T> injector) {
      injectors.put(format, injector);
    }

    public <T> void register(Format<T> format, Extractor<T> extractor) {
      extractors.put(format, extractor);
    }
  }
```

In that implementation, the fact that they're stored in a dictionary is hidden from the end user, and these signatures we can support in C# with generics. I see the benefit to giving the tracer implementors a medium-typed way to store these, but I don't see that they gain much doing this over using object (they'll still _be able to_ accept Format<T> as input, and they'll still _have to_ convert it to a strongly typed object if anyone wants to read that item).

Since storing them in a dictionary like this is an implementation detail of Jaeger allowing custom-input injectors and not a publicly exposed functionality of OpenTracing, I'd suggest that this is leaking the abstraction.

That said, it's rather pedantic and academic, so I'm okay with it, it just makes my spidey sense tingle :)

cwe1ss:
Given that it's possible to use `object` (I didn't think about that before...) and to ensure that we're as close to Java as possible, I'll close this for now. Might be something we want to improve later.

PR #70, opened 2018-02-27
Initial port of opentracing-examples from Java.

carlosalberto:
Initial import of the examples from `opentracing-java`.

Summary:

* Left out examples that are not a good fit for the C# ecosystem overall (actors, usage of `AutoFinishScopeManager`, etc).
* Used `Task`/`async` everywhere, but in `ClientServer`, where I used an actual `Thread` (this was done to exercise `AsyncLocalScopeManager`).

Please review and provide any feedback. I haven't worked full time on C# for some time now and may be losing something API/style-wise ;)

@cwe1ss


examples/OpenTracing.Examples/ActiveSpanReplacement/ActiveSpanReplacementTest.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                    // Restore the span for this task and create a subspan
+                    using (IScope subTaskScope = tracer.BuildSpan("subtask").StartActive(true))
+                    {
+                    }
+                }
+            });
+        }
+    }
+}
```
cwe1ss: Every file should end with a new-line. There's plugins for VS Code and VS 2015 (VS 2017 does it out of the box) that pick up these settings from the .editorconfig file (see #60 )

examples/OpenTracing.Examples/CommonRequestHandler/RequestHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+                spanBuilder.AsChildOf(parentContext);
+            }
+
+            context["span"] = spanBuilder.Start();
+        }
+
+        public void AfterResponse(Object response, Context context)
+        {
+            Object spanObject = context["span"];
+            if (spanObject is ISpan)
```
cwe1ss: new C# syntax: `if (context["span"] is ISpan span) { span.Finish(); }`

examples/OpenTracing.Examples/ActiveSpanReplacement/ActiveSpanReplacementTest.cs
carlosalberto: Strangely enough, VS for Mac is not picking .editorconfig up. Wonder if that's a limitation or a bug. Anyway, editing them with Vim made the trick ;)

carlosalberto:
Updated. Thanks for the review ;)

examples/OpenTracing.Examples/ActiveSpanReplacement/ActiveSpanReplacementTest.cs
cwe1ss: Seems like VS for Mac does not support it yet. You could vote for it here: https://visualstudio.uservoice.com/forums/563332-visual-studio-for-mac/suggestions/18606892-support-editorconfig

examples/OpenTracing.Examples/CommonRequestHandler/Client.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        private readonly RequestHandler _requestHandler;
+
+        public Client(RequestHandler requestHandler)
+        {
+            _requestHandler = requestHandler;
+        }
+
+        public async Task<String> Send(String message)
+        {
```
cwe1ss: This code is a bit weird IMO and looks very non-C#y :smile:

Wouldn't the Java code translate to a very simple async/await pattern like the following?

```csharp
var context = new Context();

await Task.Delay(50);
_requestHandler.BeforeRequest(message, context);
await Task.Delay(50);
_requestHandler.AfterResponse(message, context);
return $"{message}:response";
```

examples/OpenTracing.Examples/CommonRequestHandler/HandlerTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            Assert.Null(_tracer.ScopeManager.Active);
+        }
+
+        // active parent is not picked up by child.
+        [Fact]
+        public void ParentNotPickedUp()
+        {
+            using (IScope scope = _tracer.BuildSpan("parent").StartActive(finishSpanOnDispose:true))
+            {
+                var responseTask = _client.Send("no_parent");
```
cwe1ss: `string response = await _client.Send("no_parent");` (and async method signature)

examples/OpenTracing.Examples/CommonRequestHandler/HandlerTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            // Here check that there is no parent-child relation although it should be because child is
+            // created when parent is active
+            Assert.NotEqual(parent.Context.SpanId, child.ParentId);
+        }
+
+        // Solution is bad because parent is per client (we don't have better choice).
+        // Therefore all client requests will have the same parent.
+        // But if client is long living and injected/reused in different places then initial parent will not be correct.
+        [Fact]
+        public void BadSolutionToSetParent()
```
cwe1ss: async & await (see previous comments)

examples/OpenTracing.Examples/CommonRequestHandler/README.md
```diff
@@ -0,0 +1,32 @@
+# Common Request Handler example.
+
+This example shows an `ISpan` used with `RequestHandler`, which is used as a middleware (as in web frameworks) to manage a new `ISpan` per operation through its `BeforeRequest()`/`AfterResponse()` methods.
+
+Since its methods are not guaranteed to be run in the same thread, activation of such `ISpan`s is not done.
+
+```cs
+public void BeforeRequest(Object request, Context context)
+{
+    // we cannot use active span because we don't know in which thread it is executed
```
cwe1ss: Hmm.. that's no problem in C# with our `AsyncLocalScopeManager`. It correctly flows the scope, no matter on which thread the code is continued.

examples/OpenTracing.Examples/CommonRequestHandler/README.md
```diff
@@ -0,0 +1,32 @@
...
+    {
+        spanBuilder.AsChildOf(parentContext);
+    }
+
+    context["span"] = spanBuilder.Start();
+}
+
+public void AfterResponse(Object response, Context context)
+{
+    Object spanObject = context["span"];
```
cwe1ss: `ISpan span = context["span"] as ISpan;`

examples/OpenTracing.Examples/CommonRequestHandler/RequestHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Collections.Generic;
+using OpenTracing.Tag;
+
+namespace OpenTracing.Examples.CommonRequestHandler
+{
+    // One instance per Client. Executed concurrently for all requests of one client. 'beforeRequest'
+    // and 'afterResponse' are executed in different threads for one 'send'
+    public class RequestHandler
+    {
+        internal const String OperationName = "send";
```
cwe1ss: "string"

examples/OpenTracing.Examples/CommonRequestHandler/RequestHandler.cs
```diff
@@ -0,0 +1,50 @@
+using System;
+using System.Collections.Generic;
+using OpenTracing.Tag;
+
+namespace OpenTracing.Examples.CommonRequestHandler
+{
+    // One instance per Client. Executed concurrently for all requests of one client. 'beforeRequest'
```
cwe1ss: Uppercase first letter for "beforeRequest", "afterRequest", "send"

examples/OpenTracing.Examples/CommonRequestHandler/RequestHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+        }
+
+        public RequestHandler(ITracer tracer, ISpanContext parentContext)
+        {
+            _tracer = tracer;
+            _parentContext = parentContext;
+        }
+
+        public void BeforeRequest(Object request, Context context)
```
cwe1ss: "object" instead of "Object" (last comment regarding this - pls fix everywhere - same for "string" vs "String")

examples/OpenTracing.Examples/CommonRequestHandler/RequestHandler.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            // we cannot use active span because we don't know in which thread it is executed
```
cwe1ss: See previous comment - we could (and probably should) do this in C#. It should work with a `StartActive(true)` in `BeforeRequest` and a `_tracer.ScopeManager.Active.Dispose()` in `AfterResponse`.

examples/OpenTracing.Examples/LateSpanFinish/LateSpanFinishTest.cs
```diff
@@ -0,0 +1,64 @@
...
+            TestUtils.AssertSameTrace(spans);
+
+            Assert.Null(_tracer.ActiveSpan);
+        }
+
+        // Fire away a few subtasks, passing a parent ISpan whose lifetime
+        // is not tied at-all to the children.
+        // NOTE: As opposed to Java, there is not need to reactivate the parent Span,
+        // as the context is propagated by AsyncLocalScopeManager.
+        private void SubmitTasks()
```
cwe1ss: I think you could make this method return a Task and store both tasks and just return "Task.WhenAll(task1, task2)". You can then just `await` the call and don't have to call "WaitForSpanCount".

examples/OpenTracing.Examples/ListenerPerRequest/Client.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public Client(ITracer tracer)
+        {
+            _tracer = tracer;
+        }
+
+        // Async execution
+        private Task<String> Execute(String message, ResponseListener responseListener)
+        {
+            return Task.Run(() =>
```
cwe1ss: Also looks very non-C#y :smile: How about just adding a "await Task.Delay(10)" to force the async execution?

examples/OpenTracing.Examples/ListenerPerRequest/Client.cs
```diff
@@ -0,0 +1,37 @@
...
+                // send via wire and get response
+                String response = $"{message}:response";
+                responseListener.OnResponse(response);
+                return response;
+            });
+        }
+
+        public Task<String> Send(String message)
+        {
+            ISpan span = _tracer.BuildSpan("send").
```
cwe1ss: nit: "." should be on next line

examples/OpenTracing.Examples/ListenerPerRequest/ListenerTest.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace OpenTracing.Examples.ListenerPerRequest
+{
+    // Each request has own instance of ResponseListener
+    public class ListenerTest
+    {
+        private readonly MockTracer _tracer = new MockTracer();
+
+        [Fact]
+        public void test()
```
cwe1ss: async/await

examples/OpenTracing.Examples/ListenerPerRequest/ResponseListener.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace OpenTracing.Examples.ListenerPerRequest
+{
+    // Response listener per request. Executed in a thread different from 'send' thread
```
cwe1ss: "Send"

examples/OpenTracing.Examples/MultipleCallbacks/MultipleCallbacksTest.cs
```diff
@@ -0,0 +1,42 @@
...
+using static OpenTracing.Examples.TestUtils;
+
+namespace OpenTracing.Examples.MultipleCallbacks
+{
+    public class MultipleCallbacksTest
+    {
+        private readonly MockTracer _tracer = new MockTracer();
+
+        [Fact]
+        public void test()
```
cwe1ss: async/await

examples/OpenTracing.Examples/MultipleCallbacks/MultipleCallbacksTest.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            Client client = new Client(_tracer);
+            var tasks = new Task[3];
+
+            var span = _tracer.BuildSpan("parent").Start();
+            using (IScope scope = _tracer.ScopeManager.Activate(span, finishSpanOnDispose:false))
```
cwe1ss: Couldn't this also be a simple StartActive(true) due to AsyncLocal?

examples/OpenTracing.Examples/TestUtils.cs
```diff
@@ -0,0 +1,69 @@
...
+using OpenTracing.Tag;
+using Xunit;
+
+namespace OpenTracing.Examples
+{
+    public static class TestUtils
+    {
+        public static TimeSpan DefaultTimeout { get; } = TimeSpan.FromSeconds(5);
+
+        public static void WaitForSpanCount (MockTracer tracer, int spanCount, TimeSpan timeout)
```
cwe1ss: I think, ideally, this method should not be necessary if everything is `await`ed properly.

examples/OpenTracing.Examples/CommonRequestHandler/HandlerTest.cs
```diff
@@ -0,0 +1,131 @@
...
+            _client = new Client(new RequestHandler(_tracer));
+        }
+
+        [Fact]
+        public void TwoRequests()
+        {
+            var responseTask = _client.Send("message");
+            var responseTask2 = _client.Send("message2");
+
+            responseTask.Wait(DefaultTimeout);
```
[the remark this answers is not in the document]
carlosalberto: Nice, didn't know that :)

examples/OpenTracing.Examples/ActiveSpanReplacement/ActiveSpanReplacementTest.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public class ActiveSpanReplacementTest
+    {
+        private readonly MockTracer _tracer = new MockTracer();
+
+        [Fact]
+        public void test()
+        {
+            // Start an isolated task and query for its result in another task/thread
+            ISpan initialSpan = _tracer.BuildSpan("initial").Start();
```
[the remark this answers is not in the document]
carlosalberto: So the scenario to reproduce is 

1. To pass a `Span` to a task, **without** it being parent of further created `Span`s
2. Later on in the task, reactivate the passed `Span` and finish it.

Since context is propagated in C#, further created `Span`s would become children of it, and would have ended up using `IgnoreActiveSpan()` to have the same effect.

(I should do a PR against that Java example, do not activate the `Span` ;) )

examples/OpenTracing.Examples/TestUtils.cs
carlosalberto: I left it as I still need to for the ClientServer example - specially because I'm using an actual `Thread` to act as the server ;) But the rest of the examples do not use it anymore.

cwe1ss: @carlosalberto did you forget to push your changes? I don't yet see them here.

carlosalberto:
Hey @cwe1ss, thanks for the review (sorry, hadn't seen my changes hadn't hit Github ;) )

So my original idea was to 'reproduce' the examples and have them look **close** to the Java ones - but upon second thought (after your feedback), decided to go with them using `async`/`await` all the way.

IIRC, `async`/`await` and the `Task` layer overall shouldn't be used by library authors, so they will have to go to threads directly, but we can consider providing more examples for that in the future (if/as needed).

examples/OpenTracing.Examples/ListenerPerRequest/README.md
```diff
@@ -0,0 +1,23 @@
+# Listener Response example.
+
+This example shows an `ISpan` created upon a message being sent to a `Client`, and its handling along a related, **not shared** `ResponseListener` object with a `OnResponse(string response)` method to finish it.
+
+```cs
+private Task<string> Execute(string message, ResponseListener responseListener)
+{
+    return Task.Run(() =>
```
cwe1ss: Update example with latest code (await).

examples/OpenTracing.Examples/MultipleCallbacks/README.md
```diff
@@ -0,0 +1,32 @@
...
+    {
+        await Task.Delay(TimeSpan.FromMilliseconds(milliseconds));
+    }
+
+    return message + "::response";
+}
+
+// Client.Send() call time.
+var span = tracer.BuildSpan("parent").Start();
+using (IScope scope = tracer.ScopeManager.Activate(span, finishSpanOnDispose:false))
```
cwe1ss: example needs to be updated as well.

cwe1ss:
> IIRC, async/await and the Task layer overall shouldn't be used by library authors, so they will have to go to threads directly, but we can consider providing more examples for that in the future (if/as needed).

People should absolutely use `async/await` and `Task` if they work with asynchronous code. This is all over the place if you look at the code of e.g. ASP.NET Core, EF Core, .NET Core, etc. You would only need `Thread` etc. if you want to do stuff in parallel.

carlosalberto:
Hey @cwe1ss 

Updated the READMEs where needed. Thanks for that second pair of eyes ;)

Regarding the CommonRequestHandler example - since we are trying to showcase how a library using middleware-alike classes would work, I went and:

1) Created an interface called `IRequestHandler`, to make it clear this is supposed to be an external component (where OT library authors could plug in their instrumentation).
2) Relied entirely on `ScopeManager` for the creation/activation/deactivation of the `Span`, but left a note on why we don't need `Context` (as opposed to Java).

Hope that makes things better ;)

carlosalberto:
@cwe1ss Sounds fine to me! and thanks again for the feedback ;)

cwe1ss:
I'll merge this now. If anyone has more feedback, please create a new issue.

PR #75, opened 2018-03-13
StartActive() overload with finishSpanOnDispose:true

### LINKED ISSUE #73: Add a parameterless ISpanBuilder.StartActive() overload with the behavior of finishSpanOnDispose:true
`IScope StartActive(bool finishSpanOnClose);` should not the default be to close the spans on start? I'd presume this behavior should be opt-out rather than opt-in?

cwe1ss:
Resolves #73.

This is done by adding the method to the interface. I *think* this is better as it keeps everything in one place. We have to rely on tracers implementing this correctly but this shouldn't be a concern IMO (as they also have to implement all other behaviors correctly).

An alternative would be to add the method as an extension method. The advantage would be that it is non-breaking and that we could provide the implementation. However this would make the API a little bit more complex as methods would then be on interfaces and extension types. Also, NoopTracer could no longer immediately return (which probably is a negligible cost though).

Thoughts?

### DIFF
--- a/src/OpenTracing/IScopeManager.cs
+++ b/src/OpenTracing/IScopeManager.cs
@@ -14,7 +14,7 @@ namespace OpenTracing
         /// <see cref="IScope.Span"/>.
         /// <para>
         /// If there is an non-null <see cref="IScope"/>, its wrapped <see cref="ISpan"/> becomes an implicit parent of any
-        /// newly-created <see cref="ISpan"/> at <see cref="ISpanBuilder.StartActive"/> time (rather than at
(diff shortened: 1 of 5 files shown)


carlosalberto:
> An alternative would be to add the method as an extension method.

I think it's fine to put that in the main API (on the latest Cross-Language group call on Friday we talked briefly about this, and on Python we will also be providing such overload with `finish_span_on_close=True`).

cwe1ss:
@carlosalberto what's the latest status here? Can we add this overload in a next release or does it need further discussion?

cwe1ss:
As discussed in #73, this should be fine so I'll merge it. Feel free to post any concerns in #92 before we do the next release or create a new issue afterwards.

PR #77, opened 2018-03-14
Make NoopTracer types public

cwe1ss:
I want to be able to check if a given `ITracer` instance is a `NoopTracer`. My scenario is my [csharp-netcore](https://github.com/opentracing-contrib/csharp-netcore/issues/8) instrumentation library where I don't even want to start the instrumentation components if no tracer is registered. 

It's currently possible to do this via `if (tracer == NoopTracerFactory.Create()) { }` but that looks a bit weird IMO.

In opentracing-java the `Noop*` types are already public via interfaces - see e.g. [`NoopSpan`](https://github.com/opentracing/opentracing-java/blob/master/opentracing-noop/src/main/java/io/opentracing/noop/NoopSpan.java#L21). We haven't migrated these interface as it's not possible in C# to add the static `INSTANCE` variables to these interfaces.

So by making our types `public`, we would fix another deviation.

Note that the constructors are still private and the static `Instance` variables are still `internal` so the only way to get a `NoopTracer` still is a call to `NoopTracerFactory.Create`.

### DIFF
--- a/src/OpenTracing/Noop/NoopScopeManager.cs
+++ b/src/OpenTracing/Noop/NoopScopeManager.cs
@@ -3,7 +3,7 @@ namespace OpenTracing.Noop
     /// <summary>
     /// A noop (i.e., cheap-as-possible) implementation of an <see cref="IScopeManager"/>.
     /// </summary>
-    internal sealed class NoopScopeManager : IScopeManager
+    public sealed class NoopScopeManager : IScopeManager
     {
         internal static readonly NoopScopeManager Instance = new NoopScopeManager();
 
@@ -23,7 +23,7 @@ namespace OpenTracing.Noop
             return nameof(NoopScopeManager);
         }
 
-        internal sealed class NoopScope : IScope
+        public sealed class NoopScope : IScope
         {
             internal static readonly NoopScope Instance = new NoopScope();
 
--- a/src/OpenTracing/Noop/NoopSpan.cs
+++ b/src/OpenTracing/Noop/NoopSpan.cs
@@ -3,7 +3,7 @@ using System.Collections.Generic;
 
 namespace OpenTracing.Noop
 {
-    internal sealed class NoopSpan : ISpan
+    public sealed class NoopSpan : ISpan
     {
         internal static readonly NoopSpan Instance = new NoopSpan();
 
--- a/src/OpenTracing/Noop/NoopSpanBuilder.cs
+++ b/src/OpenTracing/Noop/NoopSpanBuilder.cs
@@ -2,7 +2,7 @@ using System;
 
 namespace OpenTracing.Noop
 {
-    internal sealed class NoopSpanBuilder : ISpanBuilder
+    public sealed class NoopSpanBuilder : ISpanBuilder
     {
         internal static NoopSpanBuilder Instance = new NoopSpanBuilder();
 
--- a/src/OpenTracing/Noop/NoopSpanContext.cs
+++ b/src/OpenTracing/Noop/NoopSpanContext.cs
@@ -3,7 +3,7 @@ using System.Linq;
 
 namespace OpenTracing.Noop
 {
-    internal sealed class NoopSpanContext : ISpanContext
+    public sealed class NoopSpanContext : ISpanContext
     {
         internal static readonly NoopSpanContext Instance = new NoopSpanContext();
 
--- a/src/OpenTracing/Noop/NoopTracer.cs
+++ b/src/OpenTracing/Noop/NoopTracer.cs
@@ -2,7 +2,7 @@ using OpenTracing.Propagation;
 
 namespace OpenTracing.Noop
 {
-    internal sealed class NoopTracer : ITracer
+    public sealed class NoopTracer : ITracer
     {
         internal static readonly NoopTracer Instance = new NoopTracer();
 


ndrwrbgs:
This usage violates [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle), I'd recommend finding another way to do this.

cwe1ss:
I disagree that LSP applies here. LSP is about the types themeselves and their contract. In our case, it just says that e.g. any `ITracer` implementation should do what the contract says.

Also, this is a performance sensitive library so having better performance is (a lot) more important than following patterns IMO. It's just not smart to start the instrumentation and run everything through NoopTracer if I don't have to. 

What other solution would you recommend?

MikeGoldsmith:
I agree LSP isn't of paramount importance and being able to determine if a Tracer instance is a `NoopTracer` is valuable. I also don't see the problem with exposing the NoopTracer, NoopSpanBuilder and NoopSpan as public so long as their constructors are private.

Alternatively / in addition, we could add an utility method to the NullTracer namespace that given an `ITracer` returns a bool? It could either be an extension method or directly on the NoopTracer class.

```
public static bool IsNullTracer(ITracer tracer)
{
    return tracer is NoopTracer;
}
```

cwe1ss:
> @MikeGoldsmith : Alternatively / in addition, we could add an utility method to the NullTracer namespace that given an ITracer returns a bool? It could either be an extension method or directly on the NoopTracer class.

One additional thing to consider when deciding if a `ITracer` instance is a `NoopTracer` is that the instance could also be a `GlobalTracer` with an underlying `NoopTracer`. For that reason, I'm currently already using the following helper method in my library - which is a bit hacky:
```cs
public static bool IsNoopTracer(this ITracer tracer)
{
    // This will be "if (tracer is NoopTracer)" if this PR gets merged.
    if (tracer == NoopTracerFactory.Create())
        return true;

    // There's currently no way to check the underlying tracer on the instance 
    // so we have to check the static method.
    if (tracer is GlobalTracer && !GlobalTracer.IsRegistered())
        return true;

    return false;
}
```

Would probably be nice if we could also improve the GlobalTracer check but I guess that's for a different PR.

cwe1ss:
I will make this a candidate for the next release as it's just fixing another deviation and therefore shouldn't need cross-language discussion.

I'll obviously give people more time to review though.

ndrwrbgs:
You’ve already hit yourself the problem with violating LSP, it’s more valuable than people like to give it credit. Now what happens when I’ve instrumented my code with another wrapper class that internally holds NoopTracer. I probably want the same behavior right?

The non-LSP violating way of handling this is exposing a Boolean or something that signals that this tracer implementation fulfills whatever constraints it is that you’re wanting. I’d rather follow our predecessors who got hard earned advice like LSP than stay strictly to the java APIs, if those are the choices.

That said, it honestly doesn’t matter much here, but you’ve already hit the frustration someone else is bound to hit which makes me extra cautious about this change.

ndrwrbgs:
Give me a few minutes and I’ll improve my feedback by specifying a scenario I’ve had, to give us concretes to discuss.

ndrwrbgs:
Thanks for the patience!

So the problem I hit already when I privately implemented something very similar (I'm not just pulling the advice out of nowhere, I've hit it myself and trying to save the general public from it :) ) was I had a tracer like this:

```C#
class StringBuilderTracer : ITracer {
  private readonly StringBuilder builder;
  /* Either have the underlying tracer here, or use a CompositeTracer to join this to the already existing tracer - in my case I had an ExchangeGlobalTracer method like discussed in another PR so this one had to contain the other */
  public StringBuilderTracer(StringBuilder builder) {
    this.builder = builder;
  }
  ITracer.Log(string message) {
    builder.AppendLine(message);
  }
}
```

Reason for this class: I have a developer tool that wants to expose trace messages from here-and-down (callstack wise) to the user for debugging. To do so, it captures them in-process while the real tracer sends them off to the trace system. Obviously there are other ways to do this, but unless we think no one will ever want to do any pattern like this, let's omit that discussion and focus on what's possible here.

Now with this setup, I'd probably like any code that works on NoopTracer (e.g. not starting the whole engine up) to perform the same way - I've wrapped the NoopTracer for my purposes but do not expect to have modified how it works at all.

You can image a few other scenarios that hit the same problem -- `MetricLoggingTracer`, `CompositeTracer`.


As an aside does implementation only check for NoopTracer at the first trace message, or does it do so at every trace message. I worry if a trace message from a component gets in a request to trace before I manage to set the tracer to the real instance, we'd still want subsequent messages to be traced, right?

cwe1ss:
> @ndrwrbgs Now what happens when I’ve instrumented my code with another wrapper class that internally holds NoopTracer. I probably want the same behavior right?

I'm not sure. `GlobalTracer` and `NoopTracer` are built-in features of the API and are quite special IMO (`GlobalTracer` is THE static accessor and `NoopTracer` is THE void tracer). Any other tracer will "do something" and therefore needs the instrumentation to start, right? E.g. your `StringBuilderTracer` needs active instrumentation to add traces to a StringBuilder. If someone adds e.g. a `CompositeTracer` then this tracer will need the underlying tracers as parameters so the instrumentation has to start as well. 

> @ndrwrbgs You can image a few other scenarios that hit the same problem -- MetricLoggingTracer, CompositeTracer.

Instrumentation has to start for both cases IMO.

I fully agree that checking `GlobalTracer` is very hacky though. What would be your suggested solution?

ndrwrbgs:
The best I can think of are
1) always enable your tracing code you’re teging to disable. The NOOP tracer was made efficient by design in the NOOP Case so why do you need to do more outside of it?
2) expose or use a property similar to Sampled to signal if tracing should be enabled - this is more versatile as it accounts for any tracer being disabled by config and sampling
3) make a custom flag similar to 2 for whatever implementation tracer it is that you’re building that requires this functionality. Consider it a feature of that implementation to disable all tracing rather than a feature of the API

That said, if you _disagree_ with my concerns around the risks of exposing this type please do not block for me - I will gladly defer to your judgement @cwe1ss

cwe1ss:
> @ndrwrbgs: always enable your tracing code you’re teging to disable. The NOOP tracer was made efficient by design in the NOOP Case so why do you need to do more outside of it?

The use case here is .NET's `DiagnosticSource` and how it works. If we know that we don't have a "real" tracer (= `NoopTracer`) we don't even have to subscribe to any DiagnosticSource events which makes its instrumentation almost zero-cost. Making sure we can work well with `DiagnosticSource` is very important IMO as that will probably be the main hook for almost all libraries we'll instrument.

> @ndrwrbgs: expose or use a property similar to Sampled to signal if tracing should be enabled - this is more versatile as it accounts for any tracer being disabled by config and sampling

I agree that this would be a nice way (e.g. exposed as `ITracer.IsEnabled`). This is a high-impact change though that probably needs a lot of consideration and a proper specification change. There's already some related issues for this that we could try to push forward: https://github.com/opentracing/specification/issues/8, https://github.com/opentracing/specification/issues/92

> @ndrwrbgs: That said, if you disagree with my concerns around the risks of exposing this type please do not block for me - I will gladly defer to your judgement @cwe1ss

So I'm not disagreeing with you in general, my main motivation here is parity with Java and having an easy/quick solution that doesn't have any real downsides IMO. Checking for `if (tracer is NoopTracer)` is just an optimization - if someone provides any other type it just runs through the regular logic. Replacing this with a possible future `if (tracer.IsEnabled())` would just make that check cover more scenarios.

Are you ok with proceeding with this under these circumstances?

cwe1ss:
fyi, I added my use case to https://github.com/opentracing/specification/issues/8#issuecomment-377469781

ndrwrbgs:
My concerns aren’t gone, but given our space and inability to make any quick changes to the exposed APIs I agree that this is probably the best solution available on any useful time frame - and as I agree with the motivation behind this change I withdraw my protest :)

cwe1ss:
Fair enough. 😄 Feel free to add your thoughts to the mentioned issues in the specification-repo.

I'll merge this PR around next Wednesday (after Easter) if there's no further objections.

Aaronontheweb:
I would really appreciate it if you'd merge this - I end up having to implement my own version of these types in the libraries we're consuming OpenTracing in. Would love to just use what's built into this library instead.

cwe1ss:
Thx for the reminder. I'll merge it. We still have to discuss when we'll do the next release though.

Issue #78, opened 2018-03-15
OpenTracing assembly should be signed


MikeGoldsmith:
The API assembly should be signed so it is compatible with projects that require signing. This would require a SNK file to be created and used during packaging.

Runtime exceptions are thrown if an un-signed assembly is tried to be loaded by an application that is uses signed assemblies.

MikeGoldsmith:
@cwe1ss - I'm happy to pick this one up if you could assign it to me
@tedsuo - please can you assign me repository rights so I can work on this with Christian?

cwe1ss:
I knew someone would ask that soon... :smile:

I haven't done this yet but I think it's just generating the SNK (which can be committed to git for OSS projects as it's not a security measure) and setting a few project properties.

Interested in creating a PR? If not than I can look into it in the next few days.

cwe1ss:
Oh - haven't seen your second comment. I assigned it to you, thank you.

You can also just work in your fork until you have the permissions.

MikeGoldsmith:
Thanks @cwe1ss

cwe1ss:
@MikeGoldsmith what's your ETA for this? 

I've seen that there's currently also a [OpenTracing.Signed](https://www.nuget.org/packages/OpenTracing.Signed/) package on nuget.org (with a different version number). Are you going to unlist that package once this library is signed? Would be very confusing otherwise IMO.

MikeGoldsmith:
@cwe1ss - I plan to get to it this week, I accidentally committed to it just before a week's vacation.

The OpenTracing.Signed nuget package was to assist with a packaging issue we found with the 0.10 release of OpenTracing because our Couchbase libraries are signed. Once the OpenTracing library is signed by default, we mark that package as obsolete.

PR #80, opened 2018-03-27
Add support for signing assemblies during build (#78)

MikeGoldsmith:


MikeGoldsmith:
@cwe1ss I was just typing out the same message, I've just pushed the update to include the PublicSign attribute.

MikeGoldsmith:
Tests now pass, but we need to be aware there are some consequences described [here](https://github.com/dotnet/cli/issues/6911#issuecomment-310099022).

From that link:
> However, since the binary is not actually fully signed, it cannot be installed to the GAC, cannot be used with shadow copying, and cannot be loaded in a partial trust context on full framework.

cwe1ss:
Travis CI is just a "best effort" build environment anyway. I have to deploy to NuGet from a Windows machine as we also target "net45". I'm currently deploying from my local desktop but I want to set up AppVeyor at some point.

MikeGoldsmith:
That's fair - I'm familiar with AppVeyor and can look into that integration if you like? I'll create a new issue for it.

cwe1ss:
I'll merge this as there's (hopefully) not much to object. Just comment here or create a new issue if you have any further comments.

Thanks @MikeGoldsmith !

PR #84, opened 2018-03-30
Always return NoopScope / NoopSpan for ScopeManager.Active / Tracer.ActiveSpan

cwe1ss:
This brings https://github.com/opentracing/opentracing-java/pull/263 to this repository.

Description from there:
> In special circumstances, code may automatically generate spans, negating the need to always check if Tracer.activeSpan() is null. Under these circumstances, the Noop Tracer and ScopeManager will cause NPEs when they return null values when their active methods are accessed via the GlobalTracer. This
prevents code from running successfully when using Noop Tracers.
>
> This minor change allows users to depend that their code will genuinely Noop under all use cases.

We have a separate `NoopTests` class in our repo so I had to adjust these tests. I also changed everything to `AssertSame(Noop*.Instance, ...)` instead of `Assert.IsType<Noop*>(...)` as this is a better check for our use case IMO.

As I changed these tests as well, I'll keep this PR open for a few days.

### DIFF
--- a/src/OpenTracing/Noop/NoopScopeManager.cs
+++ b/src/OpenTracing/Noop/NoopScopeManager.cs
@@ -7,7 +7,7 @@ namespace OpenTracing.Noop
     {
         internal static readonly NoopScopeManager Instance = new NoopScopeManager();
 
-        public IScope Active => null;
+        public IScope Active => NoopScope.Instance;
 
         private NoopScopeManager()
         {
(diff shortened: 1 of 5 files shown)


cwe1ss:
Please file any objections until May 2nd - I'll merge afterwards if there's none.

cwe1ss:
Found my own concerns 😄 and filed them here https://github.com/opentracing/specification/issues/116. I'll put this on hold until that issue is resolved.

PR #87, opened 2018-04-06
Rename examples to testbed.

carlosalberto:
Renamed `examples` to `testbed`, along the namespace name, to align to https://github.com/opentracing/opentracing-java/issues/185


MikeGoldsmith:
Was this really necessary? I understand the need to get the core API to match the Java implementation as close as possible but I think this is superfluous, with Examples being a more accurate description of what is there.

cwe1ss:
Hmm. We had to change the examples/testbed code anyway because of async/await so maybe our code is closer to being actual "examples" than in Java?!

I'm ok with both so feel free to merge/not merge as you like!

PR #96, opened 2018-05-01
Implement "Trace Identifiers" RFC

cwe1ss:
This implements the new ["Trace Identifiers RFC"](https://github.com/opentracing/specification/blob/master/rfc/trace_identifiers.md) by adding `TraceId` & `SpanId` members to `ISpanContext`.

The interesting thing to discuss here is the naming collision with existing tracer properties - e.g. `MockTracer` already had properties with the same names.

In this PR I changed the `MockTracer` properties to `string` but C# also has the possibility to "explicitly implement an interface" which would make it possible to keep the properties as `long`. This would look like this:

```csharp
public sealed class MockSpanContext : ISpanContext
{
    // The tracer has its own TraceId property with whatever type it wants
    public long TraceId { get; }
    
    // Whenever someone works with the `ISpanContext` interface, 
    // he can still access the `string` version of TraceId.
    string ISpanContext.TraceId => TraceId.ToString(CultureInfo.InvariantCulture);
}

// Usage examples:

// Type is string because we have ISpanContext.
ISpanContext context = span.Context;
string traceId = context.TraceId;

// Type is long because we use the concrete type.
MockSpanContext context = (MockSpanContext)span.Context;
long traceId = context.TraceId;
```

Obviously, this is a bit confusing as it's quite easy to get the "wrong" type depending on what one is doing. 

As `MockTracer` is used for unit tests where checking these properties is often necessary, I thought that changing the types to `string` makes this much less confusing.

For a real tracer, just explicitly implementing the interface and keeping the original types for its properties might be the best option though.

### DIFF
--- a/examples/OpenTracing.Examples/ActiveSpanReplacement/ActiveSpanReplacementTest.cs
+++ b/examples/OpenTracing.Examples/ActiveSpanReplacement/ActiveSpanReplacementTest.cs
@@ -33,7 +33,7 @@ namespace OpenTracing.Examples.ActiveSpanReplacement
 
             // initial task is not related in any way to those two tasks
             Assert.NotEqual(spans[0].Context.TraceId, spans[1].Context.TraceId);
-            Assert.Equal(0, spans[0].ParentId);
+            Assert.Null(spans[0].ParentId);
 
             Assert.Null(_tracer.ScopeManager.Active);
         }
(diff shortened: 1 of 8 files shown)


cwe1ss:
@opentracing/opentracing-c-maintainers feedback/approval please so that we can include this in the next release.

PR #99, opened 2018-05-14
AsyncLocalScopeManager isolation

### LINKED ISSUE #93: AsyncLocalScopeManager uses static for an instance variable
```
private static AsyncLocal<IScope> s_current = new AsyncLocal<IScope>();
```

If the user is using 2 tracers at the same time, both using an instance of AsyncLocalScopeManager underlying, then there is only 1 `s_current` shared between both of these tracers. I believe this was meant to be a private readonly variable rather than static.

It's minor, but conceivable that composite tracers may hit this problem.

ndrwrbgs:
Fixes #93. https://github.com/opentracing/opentracing-csharp/commit/aef788dda7b577ffce25714348076d5b0cdde0d0 demonstrates a test that fails before the change and passes after it.

**Problem:** Separate instances of `AsyncLocalScopeManager` will overwrite each other's data.
**Fix:** Use non-static `AsyncLocal` in the non-static `AsyncLocalScopeManager` class, and use `Guid` instead of `AppDomain.Id` for the pre4.6 implementation.

### DIFF
--- a/test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
+++ b/test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
@@ -4,6 +4,8 @@ using Xunit;
 
 namespace OpenTracing.Tests.Util
 {
+    using OpenTracing.Mock;
+
     public class AsyncLocalScopeManagerTests
     {
         private AsyncLocalScopeManager _source;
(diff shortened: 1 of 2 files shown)


test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
```diff
@@ -4,6 +4,8 @@
 
 namespace OpenTracing.Tests.Util
 {
+    using OpenTracing.Mock;
```
cwe1ss: usings outside of namespace please.

test/OpenTracing.Tests/Util/AsyncLocalScopeManagerTests.cs
```diff
@@ -14,6 +16,21 @@ public AsyncLocalScopeManagerTests()
             _source = new AsyncLocalScopeManager();
         }
 
+        [Fact]
+        public void InstancesSouldNotShareData()
```
cwe1ss: typo: "Sould" -> "Should"

cwe1ss:
I fixed the code style issues so that I can add it to the upcoming RC release. Thanks again for the PR!

Issue #92, opened 2018-04-26
v0.12.0 Release Plan


cwe1ss:
Milestone details: https://github.com/opentracing/opentracing-csharp/milestone/5

Relevant merged changes so far:
* #74 Improved XML docs
* #80 OpenTracing assembly should be signed
* #83 Updating MockTracer to use AsyncLocalScopeManager

These have been small non-breaking changes. We could do a servicing release with them but I'd prefer having fewer big releases to make sure people don't have to update their dependencies too often. For that reason, I'd like to also get some other important issues done. 

My candidates would be:
* #72 Added SetTag and WithTag for use with OpenTracing.Tag.Tags
* #75 StartActive() overload with finishSpanOnDispose:true
* #91 Use IEnumerable<KeyValuePair<string, object>> for Log() methods (again)

Would be great to get some more feedback/approval on these issues to get them done asap.

Other open questions:
* Do we already want to include an implementation for the "Trace Identifiers" RFC? If we add any of my proposed candidate issues, we will break tracers so we might as well already include this.
* Should we wait for the next Java release to have a synchronized release? There have been some commits. Not sure what else will be in their next release and when it will be.

What's your opinions on the frequency of releases? Anything else that should be included?

MikeGoldsmith:
Hey @cwe1ss - thanks for collating the release milestone. I think what you have proposed all makes sense and I agree including the additional three commits would be beneficial.

Regarding your open questions:
- It looks like #72 and #91 are breaking changes so if Trace Identifiers RFC is signed off, I think it would be a good time to add them in 0.12 too
- I don't think we need to wait for the Java release. .NET caught up with Java in 0.11 release with a big focus on reducing inconsistencies, if they have not been able to progress on roadmap issues it shouldn't stop this project moving forward.

cwe1ss:
I merged the other PRs and created the "0.12.0-rc1" release candidate!

Release notes: https://github.com/opentracing/opentracing-csharp/releases/tag/0.12.0-rc1
NuGet package: https://www.nuget.org/packages/OpenTracing/0.12.0-rc1

Please give it a try! I'd like to do the final release at the end of next week if there's no objections.

PS: Not sure yet if we should include #84 as well?! There's no definitive answer yet in that linked issue.

MikeGoldsmith:
Thanks for all your work on this @cwe1ss

cwe1ss:
No objections so far, so I'll go ahead and release it NOW.

[...]

PR #100, opened 2018-05-19
Calculate version number based on environment/github tag

cwe1ss:
I'd like to do a RC-release with the upcoming changes. However, this is not possible with the current build script as it will always do a VersionSuffix-less release if there's a git tag.

With these changes, I introduced a separate version.props file that now contains all of the version number logic (similar to ASP.NET Core projects). I'm copying the description with the new logic from that file here for further details:

>These settings ensure that we don't have to manually increment version numbers after every release!
>
>If there's an AppVeyor tag, we always build with that tag name as the Version.
This overwrites any VersionPrefix and VersionSuffix so the tag MUST have a SemVer-compatible name!
>
>Any non-official (non-tagged) build uses a fixed (high) VersionPrefix and an incrementing suffix.
This ensures that people who want to reference a package from a regular build can do so by specifying
the complete version (e.g. 99.99.99-b0014 or 99.99.99-*) which will always
be higher than any official release and therefore will always take priority.

The thing to discuss might be the "99.99.99" version for local / regular CI builds. I have this trick from the XUnit project. The advantage with this is that we will never have to bump the version manually again (which is really easy to forget).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,14 +12,7 @@ environment:
     DOTNET_CLI_TELEMETRY_OPTOUT: 1
 
 build_script:
-  - ps: |
-        if ($env:APPVEYOR_REPO_TAG -eq $true) {
-            $VersionSuffix = ""
-        }
-        else {
-            $VersionSuffix = "ci" + $env:APPVEYOR_BUILD_NUMBER.PadLeft(4, "0")
-        }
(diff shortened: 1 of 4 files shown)


cwe1ss:
I will merge this now so that I can create the RC release. Please comment here (if that's possible), in #92 or create a new issue if you have any concerns with this.

Issue #113, opened 2018-11-26
GlobalTracer uses NoopTracer type check to determine if a Tracer has been registered


MikeGoldsmith:
When using the GlobalTracer, I've found that it uses a type check against the configured tracer to determine if a tracer has been registered. 

https://github.com/opentracing/opentracing-csharp/blob/master/src/OpenTracing/Util/GlobalTracer.cs#L62

This causes a problem for us because we wish to set a tracer in our library if one has not already been configured globally. However, if the application intentionally sets a NoopTracer, we would see that as a valid scenario to use our Tracer.

I think an alternative solution would be to use an internal bool of the GlobalTracer to indicate when it has been set instead of relying on that type check.

MikeGoldsmith:
FYI this is common in other implementations too:

Java > https://github.com/opentracing/opentracing-java/blob/master/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java#L96

austinlparker:
The internal bool idea seems fine to me. Do you want to submit a PR?

carlosalberto:
@MikeGoldsmith Was going to mention the same (and think we could use a bool field for it, too ;) )

MikeGoldsmith:
Yeah, I will create a change for both .NET and Java. 

Note: when checking the other implementations, some didn't utilise a registered method at all to prevent re-registering a tracer. eg Go, Node

carlosalberto:
Python doesn't check it btw, but that was on purpose ;)

PR #114, opened 2018-11-27
Add bool to manage whether a tracer has been registered in GlobalTracer

MikeGoldsmith:
- add s_isRegistered bool
- set s_isregistered when Registering a tracer
- add unit test to verify registering a NoopTracer set the s_isRegistered bool

- move ResetTracer logic to GlobalTracer as internal function instead of test utility for easier reusability

Addresses #113


src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -82,12 +98,16 @@ public static void Register(ITracer tracer)
             lock (s_lock)
             {
                 if (tracer == s_instance._tracer)
+                {
+                    _isRegistered = true;
```
carlosalberto: Any reason why this is required?  The s_instance._tracer & _isRegistered assignation below works atomically, no?

carlosalberto:
Overall nice, I like (with the only doubt mentioned in the code ;) )

My personal request would be to not merge it till we have it working/approved in Java as well. @MikeGoldsmith if you are low on cycles, I can cook the PR for Java - else, feel free to cook it as well :)

src/OpenTracing/Util/GlobalTracer.cs
MikeGoldsmith: We need this because the default tracer is a NoopTracer, so if you intentionally set a NoopTracer we need to set the s_isRegistered bool to true too.

MikeGoldsmith:
Thanks @carlosalberto and @austinlparker - I agree Java should be approved before merging to ensure we're consistent.

So we also plan changes in the other implementations to track when a tracer has been registered and prevent subsequent calls? ie you said Python doesn't do that check and I don't think Go does either

src/OpenTracing/Util/GlobalTracer.cs
carlosalberto: Right, right. Had forgotten this detail ;)

carlosalberto:
@MikeGoldsmith So IIRC it was discussed that languages that may likely have a collision initializing the global tracer (from different threads at startup time) should have this check. Java and C# fall into this category - and for Python this is not often the case, so it was decided we wouldn't do this until truly required (the same applies to Javascript?).

I think Java + C# is a good start for now, in any case ;)

MikeGoldsmith:
After seeing the case of keeping the Tracer reset logic in the GlobalTracerTestUtil in Java, I've re-added it back here too.

MikeGoldsmith:
@austinlparker @carlosalberto - not sure why AppVeyor is failing with unit tests. They all successfully pass for me locally.

austinlparker:
I think I need different repo permissions to be able to re-run the appveyor job, lemme look into it.

MikeGoldsmith:
@austinlparker had any luck with getting AppVeyor permissions to re-run the CI?

austinlparker:
Ah, sorry, completely slipped my mind. Let me check today or I'll bring it up tomorrow at the OTSC call.

PR #115, opened 2018-12-20
Fix minor spelling in ITracer

nengberg:
I noticed a minor spelling that I thought I should fix as an early christmas gift and as a thank you for this package! 🎅


MikeGoldsmith:
Thanks!

PR #116, opened 2019-01-16
Add a binary propagator format option

austinlparker:
This adds a builtin Binary Propagator format for consumers of the API.

### DIFF
--- a/examples/OpenTracing.Examples/LateSpanFinish/LateSpanFinishTest.cs
+++ b/examples/OpenTracing.Examples/LateSpanFinish/LateSpanFinishTest.cs
@@ -46,7 +46,7 @@ namespace OpenTracing.Examples.LateSpanFinish
             {
                 using (IScope childScope1 = _tracer.BuildSpan("task1").StartActive(finishSpanOnDispose:true))
                 {
-                    await Task.Delay(55);
+                    await Task.Delay(100);
                 }
             });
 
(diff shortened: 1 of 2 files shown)


examples/OpenTracing.Examples/LateSpanFinish/LateSpanFinishTest.cs
```diff
@@ -46,15 +46,15 @@ private Task SubmitTasks()
             {
                 using (IScope childScope1 = _tracer.BuildSpan("task1").StartActive(finishSpanOnDispose:true))
                 {
-                    await Task.Delay(55);
+                    await Task.Delay(100);
```
carlosalberto: Are these changes related to the new binary propagator?

austinlparker: No, they're adjusting test timings because we keep having PR's fail. Admittedly it's not great in this PR but I didn't want to make a separate PR and rebase just to get it.

carlosalberto:
So I think this is a good addition overall - although some feedback given to the `Binary` stuff in Java can be applied here, regarding `Stream` having members that are not related to the current scenario (`CanRead`, `CanWrite`, `CopyTo`, etc).

I'd personally be up for adding this with a note on it being experimental, so people know it could change in the next versions (which actually happened to the Java side, when we changed it from using `ByteBuffer` to the `Binary wrapper ;) )

austinlparker:
I was thinking about that, but I'm also not entirely sure how many of the problems that people ran into on Java are going to be a problem here. I like the experimental idea though.

PR #117, opened 2019-01-27
Fix minor documentation comment issues

gregkalapos:
I opened the repo with JetBrains Rider and it marked a couple of things in documentation comments with error.

Mainly 2 things:

- Missing generic parameter
- Missing namespace in `MockSpan.cs` without `using`

I think based on the specs those are correct and VS has no problem with it, still with this PR 1) Rider shows no errors anymore and 2) I see that the docs almost always use generic parameters, except in the cases this PR fixes, so this'd add consistency.


MikeGoldsmith:
Yup, looks good with me too. Thanks @gregkalapos!

gregkalapos:
Thx @cwe1ss @MikeGoldsmith!

Issue #118, opened 2019-02-27
Accessing GlobalTracer.IsRegistered is not thread-safe 


christoph-neumann-schema-de:
## Problem - what blocks you from solving the requirement?

In GlobalTracer.cs the writing to s_isRegistered is thread-save via lock(s_lock).
Problem: the accessor is not thread-safe:

public static bool IsRegistered() {
    return s_isRegistered;
}

Consequence:
I would like to initialize the GlobalTracer with something like that:
if (!OpenTracing.Util.GlobalTracer.IsRegistered())
    OpenTracing.Util.GlobalTracer.Register(Tracer);

This would not be thread-safe. It is possible that two threads get IsRegistered:=false, in parallel, and then the Register happends in parallel, which throws an unexpected exception for one of both threads.

## Proposal - what do you suggest to solve the problem or improve the existing situation?

Please make reading s_isRegistered as thread-safe as writing it.

        public static bool IsRegistered() {
            lock (s_lock);
                 return s_isRegistered;
        }
		

## Relevance:
The necessary contention is easily reached with just a few (concurrent!) NUnit Test Assemblies. I got it almost at once when I started Tracing-enabled testing.

The proposed solution would not eliminate the problem to 100%. Because Thread 1 and Thread 2 might both be executing IsRegistered and might still be scheduled between both lines if...IsReg and Reg, ending in an Exception on one side due to a race condition. However, adding the lock reduces the problem. After the patch a far higher contention would be needed.

In addition, I have locked my own threads arround the If..IsReg and Reg lines, too, to be 100% sure.
Still, I noticed the delta on your side of the code and wanted to give you this hint.

MikeGoldsmith:
Hi @christoph-neumann-schema-de 

Thanks for the interesting race condition, I'll take a look soon and see how best to move forward.

MirzaMerdovic:
Sorry, if I am a bit off topic here, but why is: [Register method not pure](https://github.com/opentracing/opentracing-csharp/blob/c534179959d4a4160e06456cc70e609f827952e8/src/OpenTracing/Util/GlobalTracer.cs#L92). 
This way we have to first call _IsRegistered()_ to check and then call: _Register()_ and since this is not thread safe it can easily happen that the call fails.

MikeGoldsmith:
@MirzaMerdovic we've added a `RegisterIfAbsent` method in #119 to make this more consistent and predictable.

MirzaMerdovic:
[...]

PR #119, opened 2019-03-04
Improve thread-safety in GlobalTracer.IsRegistered

MikeGoldsmith:
Addresses #118 by wrapping `GlobalTracer.IsRegistered()` in a lock statement.

I've also implemented `GlobalTracer.RegisterIfAbsent(tracer)` that returns a bool to indicate if the registration was successful or not. This was copied from [opentracing-java](https://github.com/opentracing/opentracing-java/blob/master/opentracing-util/src/main/java/io/opentracing/util/GlobalTracer.java#L116).

### DIFF
--- a/src/OpenTracing/Util/GlobalTracer.cs
+++ b/src/OpenTracing/Util/GlobalTracer.cs
@@ -61,7 +61,10 @@ namespace OpenTracing.Util
         /// <returns>Whether a tracer has been registered.</returns>
         public static bool IsRegistered()
         {
-            return s_isRegistered;
+            lock (s_lock)
+            {
+                return s_isRegistered;
+            }
         }
 
         /// <summary>
--- a/src/OpenTracing/Util/GlobalTracer.cs
+++ b/src/OpenTracing/Util/GlobalTracer.cs
@@ -61,10 +61,7 @@ namespace OpenTracing.Util
         /// <returns>Whether a tracer has been registered.</returns>
         public static bool IsRegistered()
         {
-            lock (s_lock)
-            {
-                return s_isRegistered;
-            }
+            return s_isRegistered;
         }
 
         /// <summary>
--- a/src/OpenTracing/Util/GlobalTracer.cs
+++ b/src/OpenTracing/Util/GlobalTracer.cs
@@ -68,6 +68,44 @@ namespace OpenTracing.Util
         }
 
         /// <summary>
+        /// Attempts to register a <see cref="ITracer"/> to back the behaviour of the global tracer (<see cref="Instance"/>).
+        /// <para>
+        /// Registration is a one-time operation, subsequent attempts do not change the registered global tracer.
+        /// </para>
+        /// </summary>
+        /// <param name="tracer">The tracer to use as the global tracer.</param>
+        /// <returns>True if the tracer was successfully registered, otherwise false.</returns>
+        public static bool RegisterIfAbsent(ITracer tracer)
+        {
+            if (tracer == null)
+                throw new ArgumentNullException(nameof(tracer), "Cannot register GlobalTracer <null>.");
+
+            if (tracer is GlobalTracer)
+                throw new ArgumentException("Attempted to register the GlobalTracer as delegate of itself.", nameof(tracer));
+
+            lock (s_lock)
+            {
+                // if re-registering the same tracer, just return true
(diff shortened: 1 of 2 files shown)


src/OpenTracing/Util/GlobalTracer.cs
```diff
@@ -61,19 +61,21 @@ public sealed class GlobalTracer : ITracer
         /// <returns>Whether a tracer has been registered.</returns>
         public static bool IsRegistered()
         {
-            return s_isRegistered;
+            lock (s_lock)
```
johnduhart: I don't think this actually solves anything, at worst it means checking is a Tracer has been registered now incurs taking a lock.

In the example code given in the issue:

```csharp
if (!OpenTracing.Util.GlobalTracer.IsRegistered())
    OpenTracing.Util.GlobalTracer.Register(Tracer);
```

Even if the lock is present, this code is still not thread safe as another thread can take the lock and register a tracer.

austinlparker: Are you sure? It seems that since `Register` is calling `RegisterIfAbsent` which is under a lock, only one of them is actually going to get it.

johnduhart: The `RegisterIfAbsent` seems fine and should stay, however I don't see the value in including a lock on the `IsRegistered` call.

austinlparker: Yeah, now that you mention it, I'm questioning it as well. @MikeGoldsmith, thoughts?

MikeGoldsmith: I think the original issue may have lead me to add that, on review I too don't think it's necessary.

The benefit it gives is that if a tracer is being registered, a call to see if a tracer has been registered will wait until that action completes (whether successfully or not).

MirzaMerdovic: Sorry for butting in here, but GlobalTracer is just a singlton if I am not mistaken, so why don't you use Lazy<T> initialization for "s_instance"? You'll get: lazy initialization and you can also use the "isThreadSafe" parameter to make the object thread safe?

johnduhart: @MirzaMerdovic `GlobalTracer` isn't really a singleton per-se, as library users can register an implementation of `ITracer` to be accessible at `GlobalTracer.Instance`. I don't see how `Lazy<T>()` could achieve that.

MirzaMerdovic: True, my mistake haven't figured out the code properly. Lazy instance couldn't ensure thread safety because the "s_isRegistered" would remain unsafe.
Sorry for troubling you guys.

MikeGoldsmith: @MirzaMerdovic no problem adding your opinion. As a weird side note, a `Lazy<T>` value may be thread-safe, but the initialiser method is not.

I do agree with @johnduhart, a `Lazy<T>` won't help in this circumstance.

I can remove the `lock` around returning the * s_isRegistered*. The only value it adds is it will wait to to return if another thread is actively trying to register a Tracer but with the preferred method of `RegisterIfAbsent`, it doesn't really matter.

MikeGoldsmith: @austinlparker what is your preference here? Returning * s_isRegistered` with or without the `lock`?

austinlparker: I think without the lock makes more sense.

MikeGoldsmith:
@austinlparker @johnduhart - I've removed the lock within the `IsRegistered` method.

Any other suggestions or comments?

austinlparker:
Nope, looks good to me.

johnduhart:
👍

brunosantosfarfetch:
When are you planning to have a version with these changes? 0.12.0 doesn´t have them.

Nothing major but the RegisterIfAbsent would reduce some `GlobalTracer.IsRegistered(..)` and improve overall code readability.

austinlparker:
Ah, I don't have a great answer for you right now. Been wrapped up in a bunch of OpenTelemetry stuff, but I can add getting a new release out to my to-do list.

brunosantosfarfetch:
It's not a super urgent feature but it would be nice to have it when you have the time.

